PR #49, opened 2026-04-26
feat(release-gate): clean-cluster helm install smoke test (#44)

### LINKED ISSUE #44: Release-gate: clean-cluster helm install smoke test
## Problem

No release should publish stable if `helm install` on a clean cluster against the documented values does not reach `/readyz` 200. The v1.1.8 Debian route panic at startup would have been caught by this gate before the release tag promoted.

## Change

Add a release-gate job that:

1. Boots a fresh ARC namespace (or k3d for local).
2. Runs `helm install <release> <chart> -f values-production.yaml` against the build under test.
3. Polls `/readyz` for up to 120s.
4. Fails the release-gate if the pod does not go ready in that window.

## Acceptance criteria

- New release-gate job exists, runs in <5 min on ARC.
- Verified by deliberately introducing a startup panic on a test branch and confirming the gate fails the release.

## Tracking

Phase 1 of [Hardening Core](https://github.com/orgs/artifact-keeper/projects/2).

brandonrc:
## Summary

Adds a `clean-install-smoke` job to the release-gate workflow that performs a clean-cluster `helm install` against the documented `values-production.yaml` and verifies the backend reaches Ready and serves `/readyz` 200 within 120s.

**What this catches that v1.1.8 missed:** the v1.1.8 Debian route panic crashed the backend at startup. The pod went `CrashLoopBackOff`, the deployment rollout never completed, but the release tag promoted anyway because no gate validated a fresh install. This job watches both `kubectl rollout status deployment/<release>-backend` (catches `CrashLoopBackOff`) and a `/readyz` probe from inside the namespace (catches "process running but routes never registered"), so either failure mode trips the gate.

The job:
- Boots a fresh `smoke-<run-id>` namespace, isolated from the existing `deploy`-managed namespace.
- Runs `helm install ak-smoke <chart> -f values-production.yaml` with the chart pulled from `artifact-keeper-iac@main`. Test-only secrets (`ADMIN_PASSWORD`, `secrets.jwtSecret`, postgres password) are passed via `--set` overrides; nothing sensitive is committed. External dependencies that the smoke can't satisfy (RDS, AWS Secrets Manager, ingress, ServiceMonitor) are disabled via overrides while keeping the chart on its documented `values-production.yaml` code path.
- Polls `kubectl rollout status` for up to 120s, then polls `/readyz` from a throwaway curl pod for another 120s.
- Tears down its own namespace cleanly via a trap on EXIT, capturing pod state, events, and logs to the workflow artifact bundle on failure.

It runs independently of the existing `deploy` job (parallel, not sequential) so a startup crash fails the gate fast, before the long-running format/auth/stress matrix burns runner time on a broken build. Failure wires into `collect-results` via `needs[]`, and the existing `contains(needs.*.result, 'failure')` gate check picks it up automatically.

Phase 1 of [Hardening Core](https://github.com/orgs/artifact-keeper/projects/2). Closes #44.

## Test Checklist

- [ ] Unit tests added/updated
- [ ] Integration tests added/updated (if applicable)
- [x] E2E tests added/updated (this PR is new release-gate infrastructure)
- [x] Manually tested locally (script syntax-checked with `bash -n` and `shellcheck`; workflow YAML validated with `python -c "import yaml; yaml.safe_load(...)"`)
- [x] No regressions in existing tests (new job is added, existing matrix is unchanged)

## API Changes

- [ ] New endpoints have `#[utoipa::path]` annotations
- [ ] Request/response types have `#[derive(ToSchema)]`
- [ ] OpenAPI spec validates: `cargo test --lib test_openapi_spec_is_valid`
- [ ] Migration is reversible (if applicable)
- [x] N/A - no API changes

## Regression Test

N/A - this is not a bug fix. This PR _is_ the regression-test infrastructure that would have caught v1.1.8.

## How to verify

1. Trigger the release-gate workflow with a known-good `:dev` backend tag and confirm `clean-install-smoke` passes in under 5 minutes.
2. On a throwaway test branch, deliberately break startup (e.g. `panic!` in `main`) and confirm the gate fails with a non-zero rollout status and pod logs uploaded as a workflow artifact.

### DIFF
--- a/.github/workflows/release-gate.yml
+++ b/.github/workflows/release-gate.yml
@@ -62,6 +62,54 @@ env:
 
 jobs:
   # -------------------------------------------------------------------
+  # Clean-install smoke test
+  #
+  # Boots a fresh namespace, runs `helm install` against the documented
+  # values-production.yaml, and polls kubectl rollout status plus /readyz
+  # for up to 120s. This catches startup panics (e.g. the v1.1.8 Debian
+  # route panic) that crash the backend before it can ever serve traffic.
+  #
+  # Runs independently of the `deploy` job so a startup crash fails the
+  # gate fast, before the long-running matrix burns runner time on a
+  # broken build.
+  # -------------------------------------------------------------------
+  clean-install-smoke:
+    runs-on: ak-e2e-runners
+    timeout-minutes: 8
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          repository: artifact-keeper/artifact-keeper-test
+
+      - name: Install kubectl
+        uses: azure/setup-kubectl@v4
+
+      - name: Install Helm
+        uses: azure/setup-helm@v4
+
+      - name: Run clean-install smoke test
+        env:
+          BACKEND_TAG: ${{ inputs.backend_tag }}
+          WEB_TAG: ${{ inputs.web_tag }}
+          GITHUB_RUN_NUMBER: ${{ github.run_number }}
+        run: |
+          chmod +x scripts/clean-install-smoke.sh
+          RUN_ID="smoke-$(date +%s)-${GITHUB_RUN_NUMBER}"
+          ./scripts/clean-install-smoke.sh \
+            --run-id "${RUN_ID}" \
+            --backend-tag "${BACKEND_TAG}" \
+            --web-tag "${WEB_TAG}" \
+            --timeout 120
+
+      - name: Upload smoke diagnostics
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: clean-install-smoke-logs
+          path: /tmp/test-logs/
+          if-no-files-found: ignore
+
+  # -------------------------------------------------------------------
   # Deploy test environment
   # -------------------------------------------------------------------
   deploy:
@@ -707,7 +755,7 @@ jobs:
   # Collect results and publish summary
   # -------------------------------------------------------------------
   collect-results:
-    needs: [deploy, format-tests, security-tests, compatibility-tests, repo-tests, promotion-tests, rbac-tests, lifecycle-tests, webhook-tests, search-tests, platform-tests, auth-tests, stress-tests, resilience-tests, mesh-tests]
+    needs: [clean-install-smoke, deploy, format-tests, security-tests, compatibility-tests, repo-tests, promotion-tests, rbac-tests, lifecycle-tests, webhook-tests, search-tests, platform-tests, auth-tests, stress-tests, resilience-tests, mesh-tests]
     if: always()
     runs-on: ak-e2e-runners
     steps:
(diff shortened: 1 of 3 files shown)


brandonrc:
Pushed `5ac5de8` addressing the consolidated 4-agent senior review (DevOps Automator, Senior Developer, Reality Checker, API Tester — all returned REQUEST CHANGES on the original).

## Show-stoppers fixed

- **OpenSearch override.** Production values would render a 3-replica StatefulSet with 3× 50Gi PVCs that wouldn't be green in 120s on `local-path` (`WaitForFirstConsumer`). Backend has a hard dep on OpenSearch so we can't disable it entirely; squashed to single-node, no-PVC, 512m heap.
- **`GHCR_DOCKER_CONFIG` wiring.** Was read by the script but never set in workflow step `env:` — silent dead code. Now wired from `secrets.GHCR_DOCKER_CONFIG` with a loud `WARN` when unset so the silent path is impossible.
- **"Fast-fail" claim is now true.** `deploy needs: clean-install-smoke`, so a smoke failure short-circuits the entire matrix instead of running alongside it.
- **Timeout budget.** Backend `startupProbe` is `failureThreshold: 30 × periodSeconds: 5 = 150s`; old gate timeout was 120s. Default bumped to 300s, workflow `timeout-minutes` to 12.
- **Web rollout watched.** Web crashes are equally a broken release. `kubectl rollout status` now polled for both backend AND web.
- **Meta-test added (`.github/workflows/smoke-self-test.yml`).** Runs the gate against a nonexistent backend tag with `--expect-failure` on every PR touching the gate. If a future change silently turns the gate into a no-op (`|| true`, raised timeout, removed probe), the self-test trips. Implements the #883 contract for the gate itself.

## Should-fix changes

- **`kubectl rollout status` race.** `helm install --wait=false` returns before the API server creates Deployments. Added `wait_for_deployment_exists` (30s) before each rollout poll.
- **iac ref pinning.** New `--iac-ref` arg + `inputs.iac_ref` (default `main`). Release pipelines pass the iac tag matching the release.
- **Override drift.** Now runs `helm template --validate` before install. If a values key is renamed, this fails loudly instead of silently creating a top-level value.
- **RUN_ID collision.** Switched from `date +%s + run_number` to `github.run_id + github.run_attempt`. Always unique per attempt.
- **Argparse.** `--flag` with no value used to crash with `unbound variable`. Now emits a clear usage error.
- **`/readyz` probe rewritten.** Was spinning up a `curlimages/curl` pod per iteration (Docker Hub pull, race conditions on `--rm`). Now `kubectl exec`'s into the backend pod itself (tries wget → curl → bash `/dev/tcp` fallback). No external image pull, no per-iteration pod create.
- **Pod failure classification.** New `classify_pod_failure()` walks container/init-container `waiting.reason` so diagnostics distinguish CrashLoopBackOff (suspect the build) from ImagePullBackOff (suspect registry/secret) from Pending (suspect PVC).
- **Cleanup leaks.** `CHART_TMPDIR` is now removed in the trap. `mkdir LOGS_DIR` tolerates read-only `/tmp`. Diagnostics now capture `kubectl describe pod` (where ImagePullBackOff messages live) and `kubectl get pvc` so a failed run is triagable without `--keep-namespace`.

## Smaller cleanups
- Removed dead `SCRIPT_DIR` + `: "${SCRIPT_DIR}"` no-op
- Documented exit code 4 for `--expect-failure` violations
- Added scope-explicit comment block (what this gate IS / IS NOT)

## Verification
- `bash -n`: syntax OK
- `shellcheck scripts/clean-install-smoke.sh`: clean
- Both workflow YAMLs parse via PyYAML
- The self-test workflow provides the automated proof that the v1.1.8-class catch is real and continues to work

## Out-of-scope follow-ups (separate issues to file)
- `clean-install-smoke-with-deps` — second variant exercising Trivy/DT/edge/ingress for chart-template/wiring coverage
- `chart-upgrade-smoke` — install previous tag → `helm upgrade` to current; catches migration-on-upgrade
- Move override list into `values-smoke.yaml` in iac (single source of truth, owned by the chart)
- Periodic janitor for stale `smoke-*` namespaces and orphaned PVCs on `ak-e2e-runners`
- Action SHA pinning across the workflow (consistency follow-up)

PR #50, opened 2026-04-26
feat(security): E2E test that scans actually complete (#46)

### LINKED ISSUE #46: E2E scan flow completion
## Problem

Issue #871 reported scans stuck queued forever. We had no test that asserted scans actually complete, so the regression shipped.

## Change

Add an end-to-end test that:

1. Uploads a known package (e.g. an npm tarball with known CVEs).
2. Requests a scan via the API.
3. Polls scan status for up to 60 seconds.
4. Asserts status reaches `COMPLETED` (not `QUEUED`, not `FAILED`).

## Acceptance criteria

- Test exists in the release-gate suite.
- Runs green against `:1.1-dev` after the S3 IMDS cherry-pick lands.
- Would have caught issue #871 before tagging stable.

## Tracking

Phase 3 of [Hardening Core](https://github.com/orgs/artifact-keeper/projects/2).

brandonrc:
## Summary

Adds `tests/security/test-scan-completes-with-real-findings.sh`, a new release-gate test that catches two distinct classes of silent scan failure that have shipped to users.

Closes #46.

## What this catches

**Regression for issue artifact-keeper/artifact-keeper#871**

That issue reported scans stuck in the `QUEUED` state indefinitely after an S3 IMDS regression. The fix landed in #878 but no test exists that would have caught the original symptom. This test polls the per-artifact scan endpoint and fails explicitly if the scan is still `queued` or `pending` after the timeout, with a failure message that names #871 directly so future regressions are easy to triage.

**Regression for issue artifact-keeper/artifact-keeper#888**

That issue reports a related class of silent success: a Trivy scan reaches the `COMPLETED` terminal state even though the scanner pod is not running, so no scanner ever ran against the artifact. Asserting only on `status=COMPLETED` is therefore not sufficient.

To catch the #888 variant, the test also asserts the scan response carries scanner provenance. At least one of the following must be present in the response:

- a scanner identity field (`scanner`, `scanner_name`, `scanner_version`, `engine`, `tool.name`, `tool.version`)
- a scan timestamp (`scanned_at`, `completed_at`, `finished_at`)
- a findings or vulnerabilities array (even an empty explicit array proves the scanner emitted a structured report rather than the backend silently flipping the status field)

If `status=COMPLETED` but none of the above are present, the test fails with a `#888` reference in the failure message.

## Why two assertions instead of one

A single `status=COMPLETED` assertion would have caught #871 but not #888. A single "findings present" assertion would have caught #888 but missed the case where the scan never leaves `QUEUED` (the response would have a `findings` field carrying stale data from a prior run). Both assertions together cover the full silent-success surface that has actually shipped.

## Fixture

The test builds the fixture locally from a pinned `package.json` that declares `event-stream@3.3.6` (GHSA-mh6f-8j2x-4483, a known malicious release that Trivy and Grype both flag) as a dependency. No network calls, no `:latest` tags, fully deterministic. Resource names use `RUN_ID` so concurrent runs do not collide.

## Skip path

If the scan endpoint returns `404` or `503` consistently (six probe rounds), the test skips with a clear reason. This keeps local dev runs against a stack without a deployed scanner from flagging the test as broken. The release-gate environment must have a scanner deployed for the test to be meaningful, which it does today.

## Project tracking

Phase 3 of the [Hardening Core](https://github.com/orgs/artifact-keeper/projects/2) project.

## Test plan

- [x] `bash -n` clean
- [x] No em-dashes, no emoji, no ChatGPT-isms in test output or comments
- [x] Resource names use `RUN_ID` for idempotency
- [x] Fixture version pinned (no `:latest`)
- [x] Skip path verified for stacks without a scanner
- [ ] Runs green against `:1.1-dev` in the release-gate workflow once this PR is merged and the gate runs

## Regression test checkbox

N/A. This PR is itself a test addition, not a code fix.

### DIFF
--- a/tests/security/test-scan-completes-with-real-findings.sh
+++ b/tests/security/test-scan-completes-with-real-findings.sh
@@ -21,13 +21,13 @@
 #      pattern; the per-artifact scan endpoint is keyed by UUID).
 #   4. Triggers a vulnerability scan via POST /api/v1/security/scan with
 #      {artifact_id: uuid}. (The previous version of this test used a
-#      non-existent path-keyed endpoint and SKIPped on 404 — green forever.)
+#      non-existent path-keyed endpoint and SKIPped on 404 -- green forever.)
 #   5. Polls GET /api/v1/security/artifacts/{artifact_id}/scans for up to
 #      SCAN_TIMEOUT seconds.
 #   6. Asserts the latest scan's status is "completed" AND scanner_version
 #      is non-null AND completed_at is non-null AND error_message is empty.
 #      All four together prove that a scanner actually ran and emitted a
-#      report — catches the #888 silent-success class.
+#      report -- catches the #888 silent-success class.
 #   7. Cleans up.
 #
 # Skip semantics (load-bearing)
@@ -116,7 +116,7 @@ scanner_unavailable() {
 # reads installed `node_modules`, not just `package.json` declarations.
 # Without a lock file or installed deps, Trivy may emit zero findings for
 # this fixture. The #888 catch does NOT depend on findings being non-zero
-# — it asserts the scanner produced a report (scanner_version + completed_at
+# -- it asserts the scanner produced a report (scanner_version + completed_at
 # + clean error_message). A separate follow-up adds an OCI-image fixture
 # with a known-vulnerable base layer for findings-coverage assertions.
 # ---------------------------------------------------------------------------
@@ -182,17 +182,29 @@ fi
 # The per-artifact scan endpoint is /api/v1/security/artifacts/{artifact_id}/scans
 # and is keyed by UUID, NOT by repository path. The previous version of this
 # test polled a path-keyed endpoint that does not exist on the backend; the
-# response was always 404 and the test SKIPped on consecutive failures —
+# response was always 404 and the test SKIPped on consecutive failures --
 # silent-success on every release. Catch the bug class the test is meant to
 # catch instead.
 # ---------------------------------------------------------------------------
 
 begin_test "Resolve artifact_id for the uploaded fixture"
+# Capture epoch BEFORE the lookup so a later assertion can require that
+# the scan we observe was created after this point. This defeats the
+# auto-scan-on-upload false-pass: if a backend auto-scans on upload AND
+# completes BEFORE the explicit POST /security/scan, the test could
+# otherwise see that auto-scan's COMPLETED status without ever exercising
+# the manual trigger path.
+TEST_START_EPOCH=$(date -u +%s)
+
+# Real route: GET /api/v1/repositories/:key/artifacts/*path (no /metadata
+# suffix). Backend `repositories.rs` mounts the metadata handler on the
+# wildcard path. The previous `/metadata` suffix made the path itself
+# end in /metadata and 404'd.
 # shellcheck disable=SC2086  # CURL_TIMEOUT must word-split, per common.sh
 artifact_lookup_status=$(curl -s -o "${WORK_DIR}/artifact-resp.json" -w '%{http_code}' $CURL_TIMEOUT \
   -H "$(auth_header)" \
   -H "Accept: application/json" \
-  "${BASE_URL}/api/v1/repositories/${REPO_KEY}/artifacts/${ARTIFACT_PATH}/metadata") \
+  "${BASE_URL}/api/v1/repositories/${REPO_KEY}/artifacts/${ARTIFACT_PATH}") \
   || artifact_lookup_status="000"
 
 ARTIFACT_ID=""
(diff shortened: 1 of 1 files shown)


brandonrc:
Pushed `a1615a0` addressing the consolidated 4-agent senior review (DevOps Automator, Senior Developer, Reality Checker, API Tester — all REQUEST CHANGES).

## Show-stoppers fixed

- **Wrong endpoint (Reality Checker's smoking gun)**. The original test polled `/repositories/{key}/.../security/scan` which does not exist on the backend. After 6×404 the test SKIPped with exit 0 — green forever. Now uses the real API: `POST /api/v1/security/scan` with `{artifact_id: uuid}` and `GET /api/v1/security/artifacts/{artifact_id}/scans`. The `artifact_id` is resolved via the management API after upload.
- **Skip-on-503/404 silent-success vector**. The release-gate must fail when the scanner cannot run — that is exactly the #888 silent-success class. New `scanner_unavailable()` helper fails by default; local-dev opt-in via `ALLOW_SCANNER_SKIP=1` (the release-gate workflow must not set this).
- **Network errors (`000`) didn't count toward unavailable threshold** — a fully-unreachable backend would just poll silently for SCAN_TIMEOUT. Fixed.
- **Provenance check accepted JSON `null`**. `jq -r .scanner_version != ""` returned `"null" != ""` which is `true`. Replaced with three independent `jq -e` boolean predicates that fail on null/empty. All three must hold: `scanner_version` non-empty string, `completed_at` non-null timestamp, `error_message` null/empty.
- **60s timeout was below cold Trivy first-run** (90-120s). Bumped default to 180s.
- **No diagnostics on failure**. Scan response + scan_id now dumped to `$JUNIT_OUTPUT_DIR` for the release-gate workflow's upload-artifact step.

## Smaller fixes

- `unknown` state (jq parse failure) now distinguished from "stuck queued" so we don't misdiagnose response-shape drift as #871
- Cleanup trap combined with `setup_workdir`'s WORK_DIR cleanup instead of clobbering it
- `# shellcheck disable=SC2086` added for the intentional `CURL_TIMEOUT` word-split pattern

## Documented fixture caveat

The senior review noted Trivy's npm scanner reads installed `node_modules`, not just `package.json`. Without a lock file the fixture may emit zero findings. The **#888 catch does not depend on findings count** — it asserts scanner provenance (`scanner_version` + `completed_at` + clean `error_message`). Documented inline so a future reader doesn't try to "fix" the fixture to require findings.

## Out-of-scope follow-ups (filed separately)

- Add OCI-image fixture (e.g. alpine:3.10) for findings-coverage assertion — proves Trivy actually emits findings, not just provenance
- Sibling test that explicitly kills the Trivy pod mid-scan and asserts FAILED, not silent COMPLETED (the true #888 reproducer)
- Pre-warm Trivy DB before timing the real scan to remove cold-start flake risk
- DB-level state transition verification (assert at least one transient state observed before terminal)

## Verification
- `bash -n`: clean
- `shellcheck`: clean (SC1091 about external common.sh source is expected)

brandonrc:
Pushed `ed0b972` addressing the Round 2 review (DevOps + Reality + API Tester REQUEST CHANGES, Senior Dev APPROVE).

## Show-stoppers fixed

| Round 2 finding | Fix |
|---|---|
| `/metadata` URL doesn't exist (DevOps + Reality) | Switched to `/api/v1/repositories/{key}/artifacts/{path}` (no suffix) |
| Dead jq fallbacks `// .artifacts // .` (DevOps) | Removed; backend ships only `{items}` |
| Auto-scan-on-upload false-pass via `items[0]` (Reality) | New `TEST_START_EPOCH` + strict assertion that `scan.created_at >= TEST_START_EPOCH` |
| `scanner_version` stub vector (Reality + API Tester) | Tightened predicate: requires at least one digit, rejects `[stub, mock, none, unknown, test, fake]` |
| `network_fail_count` misses 502/504 (DevOps) | Added 502 and 504 to the counter |

## Smaller fixes
- Em-dash sweep: 11 unicode em-dashes replaced with `--` per CLAUDE.md style

## Round 2 findings deliberately deferred (with reasoning)

- **Workflow-side `kubectl logs trivy` failure step.** Belongs in `release-gate.yml` but the merged version of that file (post #49 merge) does not include the smoke gate or its trivy-logs hook. Re-introducing it here would conflict; filed as a follow-up so the workflow change can be reviewed independently.
- **Paired pass/fail CI proof.** Cannot be demonstrated until the test lands and runs against a `:1.1-dev` build. Marked unchecked in the PR test-plan; filed as follow-up.
- **Meta-protection (CODEOWNERS, self-test fixture, `ALLOW_SCANNER_SKIP` guard).** Filed separately as Hardening Core hygiene follow-ups; not blocking this fix-the-bug PR.

## Bug classes still NOT caught (documented, not blocking)
- Backend stub emitting `scanner_version="trivy-0.0.0-fake"` (passes digit-shape check) — would need scanner-pod-side cross-check
- Trivy DB stale or missing CVE-2018-16487 / GHSA-mh6f-8j2x-4483
- Mis-attributed findings (right CVE, wrong component)
- Whitespace-only `error_message`

## Verification
- `bash -n`: clean
- `shellcheck`: clean (SC1091 about external common.sh source is expected)
- 0 em-dashes remain

brandonrc:
Closing as superseded by epic #56.

After three review rounds, each catching at least one real defect, the consistent finding is that the gate's assumptions outpaced the backend's actual contract. Most damning: the backend never writes `scanner_version` to scan_results, so the central #888 catch (assert scanner provenance) cannot work without a backend change.

Splitting into the sub-tasks tracked in #56:
- Backend: populate `scanner_version` + `started_at` in `complete_scan` / `fail_scan`
- Test: realistic E2E gate using populated fields
- Meta-test: paired pass/fail evidence (#883 contract)
- Workflow diagnostics on failure
- Time-handling robustness (sub-second / fractional-second issues)
- Negative coverage (malformed artifact, scanner-down)

Three rounds of review documented the discovery and live in the PR comments. Next implementer should start from those before re-implementing.

Branch `feat/e2e-scan-completion` retained for reference; will be deleted after the backend sub-task lands and a fresh PR supersedes it.

PR #51, opened 2026-04-26
feat(security): E2E test that SBOM generation produces real components (#47)

### LINKED ISSUE #47: E2E SBOM correctness
## Problem

Issue #870 reported SBOM generation returning empty / placeholder data while the API responded `200`. We had no test that validated the SBOM body, only the status code, so the regression shipped.

## Change

Add an end-to-end test that:

1. Uploads a known npm package (e.g. `lodash@4.17.21`).
2. Requests SBOM generation.
3. Polls until generation completes.
4. Asserts the SBOM body has `components.length > 5` AND known component names appear.

## Acceptance criteria

- Test exists in the release-gate suite.
- Runs green against `:1.1-dev` after the S3 IMDS cherry-pick lands.
- Fixture pinned by version (no `latest`) so the assertion is deterministic.
- Would have caught issue #870 before tagging stable.

## Tracking

Phase 3 of [Hardening Core](https://github.com/orgs/artifact-keeper/projects/2).

brandonrc:
## Summary

Adds `tests/security/test-sbom-correctness.sh`, an end-to-end regression
test that protects against the silent-empty-SBOM symptom reported in
artifact-keeper#870. Closes #47.

The new test uploads a pinned npm fixture (`express@4.18.2`), waits for
the security scan, requests SBOM generation, and then asserts on the
**SBOM body** (not just the status code):

1. `HTTP 200` on `GET /api/v1/sbom/by-artifact/{artifact_id}`.
2. `components.length > 5`.
3. At least 2 of `[express, body-parser, qs]` appear by name.

## Why status 200 alone wasn't sufficient

Issue artifact-keeper#870 (community report from @Firjens, closed as
downstream of #871 / fixed by #878) showed the SBOM endpoint returning
status 200 with empty/fake data. The pre-existing `platform/test-sbom.sh`
suite validated structural fields (`bomFormat`, `specVersion`) against a
synthetic generic blob that legitimately has no dependencies, so a
`components: []` response looked correct.

This is the silent-success failure mode: the API contract was technically
satisfied while the actual SBOM data was wrong. The fix landed upstream
(S3 IMDS, #878), but the SBOM code path itself still swallowed the
storage error and persisted an empty record. We need a test that catches
that symptom even when the upstream cause is something we haven't seen
yet.

What each assertion catches:

| Failure mode                          | Caught by         |
| ------------------------------------- | ----------------- |
| Endpoint 5xx / down                   | Assertion 1       |
| Empty SBOM (`components: []`) at 200  | Assertion 2 (#870)|
| Fabricated/placeholder names at 200   | Assertion 3 (#870)|

## Fixture choice

`express@4.18.2` is pinned (no `latest`) for determinism. It has 30+
direct deps and ~50 transitive, so a healthy SBOM clears the `> 5`
threshold easily and the assertion is not flaky on minor scanner output
changes. We expect `express`, `body-parser`, and `qs` to appear; the
test requires at least 2 of those 3 to allow for occasional name-form
variation in scanner output (e.g. `qs@6.x` vs `qs`).

## Discovery

The script lives in `tests/security/`, which `scripts/run-suite.sh` and
`.github/workflows/release-gate.yml` already auto-discover via the
`security` suite choice. No orchestrator changes were needed.

## Test plan

- [x] `bash -n tests/security/test-sbom-correctness.sh` passes
- [x] `shellcheck` clean (only the pre-existing `$CURL_TIMEOUT` info-level
  pattern from `tests/lib/common.sh` and the SC1091 sourcing note)
- [x] Test is idempotent: all resource names use `$RUN_ID`, and an `EXIT`
  trap deletes the temporary repo
- [x] Fixture is pinned (`express@4.18.2`, no `latest`)
- [ ] Runs green against `:1.1-dev` after the S3 IMDS cherry-pick lands
  (pending CI on this PR)

## Checklist

- [x] Idempotent (uses `$RUN_ID`, cleans up via `EXIT` trap)
- [x] Sources `tests/lib/common.sh`
- [x] PASS/FAIL per section, JUnit output via `end_suite`
- [x] No em-dashes, no emoji, no ChatGPT-isms
- [N/A] Regression test for a bug fix (this is a test addition, not a code fix)

## Tracking

- Closes artifact-keeper-test#47
- Hardening Core (Phase 3): https://github.com/orgs/artifact-keeper/projects/2
- References artifact-keeper#870 (symptom this would have caught)
- References artifact-keeper#878 (the upstream fix)

### DIFF
--- a/tests/security/test-sbom-correctness.sh
+++ b/tests/security/test-sbom-correctness.sh
@@ -17,9 +17,12 @@
 #
 #   1. HTTP 200 on the SBOM read.
 #   2. components.length > 5  (catches the empty-SBOM case from #870).
-#   3. Specific known component names appear in the SBOM (catches the
+#   3. Specific known TRANSITIVE component names appear (catches the
 #      "fake placeholder data" case from #870, where the response shape was
-#      valid but the data did not correspond to the uploaded artifact).
+#      valid but the data did not correspond to the uploaded artifact). The
+#      uploaded package itself (express) is excluded from this assertion --
+#      it appears in any SBOM as the root component, so requiring it would
+#      not prove the scanner actually walked the dep tree.
 #
 # Fixture: express@4.18.2 (pinned, no `latest`). Chosen because it has a
 # deep, stable dependency tree (~30 direct deps, ~50 transitive) and is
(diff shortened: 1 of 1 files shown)


brandonrc:
Pushed `8e4c889` addressing the consolidated 4-agent senior review.

## Show-stoppers fixed

| Round 1 finding | Fix |
|---|---|
| Cleanup trap clobbered `setup_workdir`'s parent EXIT trap (Senior Dev) | Combined into single trap that does both `cleanup_repo` and `rm -rf "$WORK_DIR"` |
| `(.content.components \| length? // 0)` is wrong (Senior Dev) | Replaced with `((.content.components // []) \| length)` |
| `component_count` not numeric-validated before `-gt` (Senior Dev) | Added `is_nonneg_int()` helper, validate before every numeric comparison |
| jq filter precedence: `A or B and C` could mis-bind (Senior Dev) | Parenthesized each side of `and`/`or` in artifact-lookup filter |
| **Polling test passed on timeout (DevOps)** | Now uses a `populated` flag and FAILs explicitly on timeout exhaustion |
| `scan_resp` not validated as JSON (Senior Dev) | Pipe through `jq empty` first |
| 60s SBOM_POLL_TIMEOUT was tight for cold-start scanner (DevOps) | Bumped to 180s; SCAN_WAIT_TIMEOUT to 120s |
| Polling fallback `if $meta > 0 then $meta else $embedded` is preferential (Reality) | Replaced with `[$meta, $embedded] \| max` |
| **Assertion 3 too lenient: `express` itself was matchable (DevOps)** | Excluded `express` (root component); renamed list to `EXPECTED_TRANSITIVE_COMPONENTS`; added `cookie` so the 2-of-3 threshold actually proves dep-tree walking |

## Smaller fixes
- Added `# shellcheck disable=SC2086` for intentional `CURL_TIMEOUT` word-split
- Diagnostics dump on failure to `$JUNIT_OUTPUT_DIR`

## Out-of-scope follow-ups (Hardening Core)
- SPDX format coverage (only CycloneDX exercised today)
- OCI/Docker SBOM fixture for OS-package coverage
- PURL / license / hash field assertions
- Negative test: 1-byte upload distinguishes "no deps" from "scan failed"
- Self-pinning meta-protection (CODEOWNERS, threshold-lock comment block)
- Scanner version pin in docker-compose.local-dev.yml

## Verification
- `bash -n`: clean
- `shellcheck`: clean (SC1091 about external common.sh source is expected)
- 0 em-dashes

brandonrc:
Closing as superseded by epic #57.

Two review rounds (Senior Dev APPROVE on craft, Reality Checker NEEDS WORK, DevOps REQUEST CHANGES) found that this is the same #50-style trap: the test's assumptions outpace the backend's actual contract. Most damning, in addition to the polish items already addressed in `8e4c889`:

- Backend never enables `--list-all-pkgs` in Trivy. SBOM derived from `scan_findings` is "CVEs found in deps", not "deps". The 30+ deps express@4.18.2 declares would only show up in the SBOM if Trivy's DB had CVEs for them at test time.
- Component names persist as `"body-parser (package-lock.json)"` not `"body-parser"`. Test exact-match `. == "body-parser"` never matches.
- Test never sets `scan_on_upload`, so the backend never scans the upload at all.
- The `cve/history` "scan complete" probe returns `[]` for any unknown artifact, so the test treats `[]` as success silently.

Splitting into the sub-tasks tracked in #57:
- Backend: full-deptree SBOM (artifact-keeper#903)
- Test: enable `scan_on_upload`, real scan-completion probe, substring match, force_regenerate-every-iteration, negative control, Trivy DB pinning
- Workflow: scanner pod log capture on failure
- Meta-test: paired pass/fail evidence (#883 contract)

Two rounds of review live in the PR comments. Next implementer should start from those before re-implementing.

Branch `feat/e2e-sbom-correctness` retained for reference; will be deleted after the backend prerequisite (#903) lands and a fresh PR supersedes it.

Issue #48, opened 2026-04-26
Per-format push/pull audit and gap-fill


brandonrc:
## Problem

Format coverage in the release-gate is uneven. Some formats only have management-API tests; some have native-client smoke; docker format coverage was reported broken by community testers but no gate would have caught it.

## Change

1. Audit the existing `tests/formats/` suites and document which formats actually run a native-client push + pull cycle (not just the management API).
2. For at minimum maven, npm, pypi, docker, generic, debian: add or fix tests so each one has a small fixture push followed by a pull, asserting bytes match.
3. Wire all six into the release-gate matrix.

## Acceptance criteria

- Audit document committed under `docs/release-gate-coverage.md` or similar.
- All six listed formats have a passing push+pull test.
- Release-gate fails if any format suite fails.

## Tracking

Phase 3 of [Hardening Core](https://github.com/orgs/artifact-keeper/projects/2).

brandonrc:
Closed by PR #147 (Cluster A).

PR #52, opened 2026-04-26
feat(coverage): per-format push/pull audit and gap-fill (#48)

brandonrc:
## Summary

Audits every script under `tests/formats/` and characterizes whether it
runs a real native client (`mvn`, `npm`, `pip`, `twine`, `docker`,
`apt-get`), a curl call against the format-native HTTP endpoint, or
only the management API at `/api/v1/repositories/{key}/artifacts/...`.

The release gate had uneven coverage:

- `npm` already exercises real `npm publish` + `npm install`.
- `pypi` already exercises real `pip install`, but uses curl multipart for the upload (twine wire format only).
- `maven` only used curl PUT/GET against `/maven/{key}/...`. No `mvn` invocation.
- `docker` (`test-oci.sh`) defaults to a curl-based simulation of the OCI v2 protocol. Real `docker push`/`pull` is gated behind `FORCE_DOCKER_TESTS=true`, which is never set in the workflow.
- `generic` only exercised the management API. The format-native `/generic/{key}/{path}` endpoint that real consumers hit was untested.
- `debian` builds a real `.deb` and pushes via curl, but no real `apt-get install` smoke.

The docker gap is the one Firjen reported on the community channel: the
gate would not have caught a regression in the OCI distribution
endpoint, because no real docker client ever talks to the test deploy.

## Change

Adds four new tests under `tests/formats/`:

- `test-maven-native-client.sh`: `mvn deploy:deploy-file` to push, `mvn dependency:get` to pull, hermetic `settings.xml` in `WORK_DIR`, two separate local repos so the pull is a real network fetch. Auto-skips if `mvn` is missing.
- `test-pypi-native-client.sh`: bootstraps an isolated venv with pinned `twine==5.1.1` and `build==1.2.2`, builds a real sdist, runs real `twine upload`, runs real `pip install` against the simple index, and imports the installed module to confirm the bytes round-tripped.
- `test-docker-native-client.sh`: real `docker login`, `docker push`, `docker pull` against the in-cluster registry. Builds from `busybox:1.36.1` (pinned, no `:latest`). Skips cleanly if `docker` is absent, the daemon is unreachable, the registry hostname does not resolve, or the runner has no insecure-registries entry for an HTTP backend.
- `test-generic-native-client.sh`: curl PUT/GET against the format-native `/generic/{key}/{path}` endpoint with Basic auth, SHA256 round-trip, deeply nested paths, 404 on unknown paths, and an explicit anonymous-write rejection check.

Wires all four into both `release-gate.yml` (the gate) and
`format-tests.yml` (the reusable workflow caller). The `python` batch
now installs `python3-venv`; the `jvm` batch now installs `maven`.

The audit doc is committed at `docs/format-coverage-audit.md` with
the full inventory of every format script, what it actually does,
and what the new tests close.

Real `apt-get install` for debian is documented as blocked in the
audit, with the three concrete unblockers listed (GPG-signed Release
file, `signed-by` keyring, root or `--download-only`). It is filed as
follow-up rather than shipped here so the diff stays reviewable and
the test does not flake until all three pieces land.

References:
- Hardening Core: https://github.com/orgs/artifact-keeper/projects/2
- Symptom this catches: docker-format community report (Firjen) that no gate flagged before release.

## Coverage after this PR

5 of 6 required formats have real native-client push+pull: `maven`,
`npm`, `pypi`, `docker`, `generic`. Debian remains on
format-HTTP-only and is documented as blocked.

## Test Checklist
- [x] `bash -n` clean on all four new scripts
- [x] YAML lint clean on `release-gate.yml` and `format-tests.yml`
- [x] No em-dashes, no emoji
- [x] All resource names use `${RUN_ID}` for idempotency
- [x] All fixture versions pinned (busybox:1.36.1, twine==5.1.1, build==1.2.2, setuptools==75.6.0)
- [ ] Release-gate run validates the four new scripts on the next workflow_dispatch

## API Changes
- [x] N/A - test infrastructure only

## Regression Test
- [x] N/A - this is test addition (audit + gap-fill), not a bug fix


brandonrc:
Pushed `7023f64` addressing the consolidated 4-agent senior review.

## Show-stoppers fixed

- **Docker silent-skip in release-gate (DevOps + Reality)**. Pre-flight skips were bare `exit 0`s that wrote nothing to JUnit — the dashboard showed "no testcases" rather than the explicit reason, AND the gate stayed green when the test that justified the PR never ran.
  - Added `skip_suite(reason)` to `common.sh` that emits a JUnit `<skipped/>` testcase
  - Added `RELEASE_GATE=1` env var; when set, `skip_suite()` turns into a hard FAIL (pre-flight skip in release-gate is silent-success class)
  - `format-tests.yml` now sets `RELEASE_GATE: '1'`
  - All 5 bare `exit 0` SKIPs in `test-docker-native-client.sh` migrated to `skip_suite`

- **`docker info | grep` insecure-registries match bug (Senior Dev)**. Matched against `REGISTRY_HOSTNAME` (no port) but dockerd entries are keyed by host:port. A different port on the same hostname could pass silently. Now uses line-anchored regex matching `REGISTRY_HOST` (with port) first, falls back to hostname-only.

- **Push+pull didn't assert digest stability (API Tester)**. Now captures `PUSH_DIGEST` after push, compares with `PULL_DIGEST` after pull. Catches OCI silent-corruption (registry mutates layers, returns different digest).

- **No per-job timeout (DevOps)**. Added `timeout-minutes: 25` to format-tests.

## Helpers added (eliminate duplication noted by Senior Dev)
- `skip_suite(reason)` — JUnit-emitting pre-flight skip with `RELEASE_GATE` enforcement
- `assert_http_2xx(status, msg)` — replaces the inline `[ -ge 200 -lt 300 ] 2>/dev/null` pattern (the `2>/dev/null` was masking arithmetic failures on string "000")

## Out-of-scope follow-ups (Hardening Core)
- DinD runner image bake (maven, python3-venv, pip wheelhouse)
- Pre-pull busybox to in-cluster registry to remove Docker Hub dep
- Adopt `assert_http_2xx` across all native-client scripts
- `basic_auth_url` helper in pypi test
- Subshell `cd` fixes in maven/pypi native-client tests
- Rename `test-generic-native-client.sh` (misnamed; generic has no client)
- Backend-formats-vs-tests CI guard (anti-self-pinning)
- PR title rescope: "release-gate-required formats" not "audit"

## Verification
- `bash -n`: clean
- `shellcheck`: clean (pre-existing SC2086 in common.sh intentional)
- `yaml.safe_load`: clean

brandonrc:
Pushed `2ddf533` addressing the consolidated 3-agent fresh-eyes review (DevOps + Senior Dev + Reality Checker — all REQUEST CHANGES, unanimous on the load-bearing finding).

## The big fix

`RELEASE_GATE='1'` was set in `format-tests.yml`, but `release-gate.yml` is the actual release gate and never invokes `format-tests.yml`. All 3 reviewers caught this. Round 3:

- **Added `RELEASE_GATE: '1'` to all 13 release-gate.yml job env blocks** that run test scripts. Now `skip_suite()` correctly emits `<failure/>` + exit 1 in the gate that matters most.

## Other fixes

- **`assert_http_2xx` migration**: helper was added in Round 2 but never called. Migrated `test-generic-native-client.sh`'s two HTTP 2xx checks. The previous inline pattern `[ -ge 200 ] 2>/dev/null && [ -lt 300 ] 2>/dev/null` silently masked `000` (curl exit non-zero).
- **`auth_admin` ordering**: moved AFTER pre-flight skips in `test-docker-native-client.sh`. If `auth_admin` errored (backend down) the suite would error before `skip_suite` fired, bypassing silent-success protection (Reality Checker finding).
- **Digest comparison**: switched from `.Id` (local config-blob hash) to `RepoDigests` (registry manifest digest). Catches manifest-only mutation (mediaType drift, annotation injection) which `.Id` misses. (Senior Dev + Reality Checker)
- **Em-dash sweep**: 1 unicode em-dash in `common.sh:756` replaced with `--` per CLAUDE.md.

## Out-of-scope follow-up filed

- **artifact-keeper-iac#82**: widen DinD insecure-registries allowlist to include `test-${RUN_ID}` namespace services. Without this, the docker test pre-flight-skips on every ARC runner. With `RELEASE_GATE=1` (now correctly set in this PR) that skip becomes a hard fail — honest fail-closed behavior. The gate goes red until the iac change lands; that's the right signal.

Other follow-ups already filed:
- Refactor release-gate.yml to call format-tests.yml via `uses:` (single matrix definition)
- Pre-bake maven into runner image
- Backend-formats-vs-tests CI guard (anti-self-pinning)
- Migrate remaining native-client scripts to `assert_http_2xx`

## Verification
- `bash -n`: clean across all 3 modified shell files
- `shellcheck`: clean (pre-existing SC2086 intentional patterns kept)
- `yaml.safe_load`: clean
- 0 em-dashes remain in modified files

Issue #70, opened 2026-04-27
Epic 5: Format suites missing entirely E2E coverage (v1.1.9)


brandonrc:
Formats either with zero E2E coverage or with a backend/test mismatch. Tracked under coverage policy #66.

## Sub-tasks (1.1.x scope)

5.1 **Gradle**: aliased to Maven handler in backend (resolve_maven_repo accepts `["maven", "gradle"]`) but ZERO Gradle-specific tests. Add test-gradle.sh covering build.gradle.kts resolution and Gradle metadata conventions.
5.2 **jetbrains_plugins**: tests reference a handler that is not wired into the format registry (per agent finding). Audit and fix the handler/test mismatch.
5.3 **mlmodel**: handler unclear, tests exist. Audit.
5.4 **opkg**: no handler file, tests exist. Audit.
5.5 **p2**: no handler file, tests exist. Audit.
5.6 **vagrant**: no handler file, tests exist. Audit.

The 5.2-5.6 audit may reveal that some formats are routed through generic/proxy handlers and the tests are valid; or it may reveal that tests are testing nothing. Either way, ground truth needs to be established.

brandonrc:
Discovered while writing PR #80: `cache_ttl_secs` is stored but never consumed by proxy/cache code. Filed as backend bug artifact-keeper#908. Sub-task 6.3 will need a follow-up E2E assertion (cache actually evicts at TTL) once #908 is fixed.

brandonrc:
Correcting an error in PR #80: the agent's grep missed `backend/src/services/proxy_service.rs:284` (`get_cache_ttl_for_repo`). `cache_ttl_secs` IS consumed and drives cache eviction via `expires_at`. artifact-keeper#908 closed as not-a-bug. Sub-task 6.3 should be expanded to ALSO assert real eviction behavior (upload, wait past TTL, verify re-fetch), not just the storage roundtrip.

brandonrc:
## Status (2026-05-16)

The audit work this epic asked for landed on `main`:

- `tests/formats/AUDIT-epic-5.md` documents the format/test mismatch resolution: mlmodel, opkg, p2, vagrant, jetbrains_plugins all have `FormatHandler` trait impls in backend; the formats without dedicated HTTP handler modules (mlmodel/opkg/p2/vagrant) are reachable via the WASM proxy (`/ext/{format_key}/...`) and the generic artifact API. Tests already take those paths and skip cleanly when no plugin is loaded.
[...]

PR #80, opened 2026-04-27
feat(repos): virtual member CRUD + cache TTL E2E (#70)

brandonrc:
## Summary

Adds five E2E test scripts covering Epic 6 sub-tasks (artifact-keeper-test#70) for v1.1.9 coverage:

- **6.1** `tests/repos/test-virtual-repo-member-remove.sh` - `DELETE /api/v1/repositories/:key/members/:member_key`, plus a cascade-on-local-repo-delete section that closes **ak-wlib**
- **6.2** `tests/repos/test-virtual-repo-member-bulk-update.sh` - `PUT /api/v1/repositories/:key/members` (bulk priority update)
- **6.3** `tests/repos/test-cache-ttl-config.sh` - `GET/PUT /api/v1/repositories/:key/cache-ttl`
- **6.4** `tests/repos/test-virtual-repo-member-authz.sh` - write-scope rejection on virtual member + cache-ttl endpoints, closes **ak-0q3a**
- **6.5** `tests/repos/test-cache-ttl-eviction.sh` - end-to-end cache TTL eviction (upload, wait, re-fetch), closes **ak-qity**

### 6.1 Member remove

Creates two local repos A, B and a virtual V with both as members, then `DELETE /V/members/A` and asserts only B remains. Adds two error paths the handler explicitly returns:
- Unknown `member_key` -> 404 (handler resolves via `RepositoryService::get_by_key` which raises `AppError::NotFound`).
- Parent repo is not virtual -> 400 (`AppError::Validation` from `if virtual_repo.repo_type != Virtual`, deterministic per `backend/src/error.rs:91`).

A 6.1.d cascade section (added with this revision) creates a fresh virtual W with members [C, D], deletes local repo C directly via `DELETE /api/v1/repositories/C` (NOT via `/W/members/C`), and asserts W is left with one member (D). This catches the user-visible failure mode where revoking a local repo silently degrades unrelated virtual aggregates. The migration `backend/migrations/003_repositories.sql` defines `member_repo_id ... REFERENCES repositories(id) ON DELETE CASCADE`, and this section now exercises that FK.

### 6.2 Bulk member update

Backend `update_virtual_members` ONLY updates priorities for already-existing members. It does not add or remove members. The test reflects that contract:

1. Create V with members [A(p=1), B(p=2)].
2. `PUT { members: [{A, p=10}, {B, p=5}] }`.
3. GET returns members ordered by priority ascending, so the test asserts B precedes A and the priorities round-trip.
4. PUT with an unknown member_key -> 404 (resolution failure inside the loop).
5. PUT against a non-virtual parent -> 400.

The header explicitly documents the add/remove non-behavior so future readers don't assume PUT replaces the membership set.

### 6.3 Cache TTL CRUD

CRUD-only contract: get-after-set, upsert, boundary validation. The end-to-end eviction test now lives in 6.5 (was previously deferred as ak-qity).

What this script asserts:

- Default value (3600) when no row exists in `repository_config`.
- PUT echoes the value, GET returns the persisted value.
- Upsert: PUT a different value, GET reflects it (guards against an INSERT-without-ON-CONFLICT regression).
- Boundary validation per `validate_cache_ttl()`: 1 and 2_592_000 are accepted, 0 / 2_592_001 / -1 are rejected with 400.
- Endpoint works on both remote and local repos today (the backend doesn't gate by `repo_type`); if that changes the local-repo case will surface it via `skip`.
- Unknown repo -> 404.

### 6.4 Member + cache-ttl write-scope rejection (closes ak-0q3a)

Provisions a read-only API token via `POST /api/v1/auth/tokens` with `scopes:["read"]` (same pattern as `tests/rbac/test-scope-enforcement.sh`) and asserts the backend authz contract documented in `backend/src/api/handlers/repositories.rs`:

- `POST /:key/members` -> 403 (handler calls `require_scope("write")`)
- `DELETE /:key/members/:mk` -> 403 (handler calls `require_scope("write")`)
- `PUT /:key/members` -> 403 (handler calls `require_scope("write")`)
- `PUT /:key/cache-ttl` -> 403 (handler calls `require_scope("write")`)
- `GET /:key/cache-ttl` -> 2xx (no write scope required, positive control)
- `GET /:key/members` -> 2xx (no write scope required, positive control)

The two GET checks are positive controls so a deny-everything middleware regression cannot pass the negative checks silently. If `POST /api/v1/auth/tokens` does not return a usable token, the script calls `skip_suite` with a precise reason rather than fabricating a fake token; under `RELEASE_GATE=1` that surfaces as a hard failure (this is the silent-success class ak-0q3a was opened to close).

### 6.5 Cache TTL eviction (closes ak-qity)

Closes the eviction-path gap that test-cache-ttl-config.sh explicitly defers in its 6.3.b/c CRUD-only header. Strategy uses a controllable upstream and the simple-index URL (a stable URL whose body changes as new versions are uploaded), so the test does not depend on any X-Cache HIT/MISS header (the backend exposes only `X-Cache: STALE` today, and only on stale-while-revalidate).

Steps:
1. Create local PyPI repo U (controllable upstream).
2. Upload pkg v1.0.0 to U.
3. Create remote PyPI repo R pointing at U with `cache_ttl_seconds=2`.
4. Fetch `/pypi/R/simple/<pkg>/` to prime the cache.
5. Upload v2.0.0 to U directly (upstream now has v1 AND v2).
6. Within-TTL fetch must NOT show v2 (cache hit; TTL doing its job).
7. Wait 6s (TTL=2 + 4s buffer for scheduler/storage jitter).
8. Post-TTL fetch must show v2 (cache miss forced re-fetch).

Step 6 has a slipped-window guard: if test scheduling pushes elapsed time past the TTL before the within-TTL fetch lands, the assertion is skipped rather than made flaky; the post-TTL eviction assertion still runs. Avoids the stale-while-revalidate branch (proxy_service.rs:175) by keeping the upstream responsive throughout, so the only difference between step 6 and step 8 is elapsed time relative to the configured TTL. PyPI is used instead of generic format because `/generic/` has no proxy route wired to `proxy_service` in the current backend.

## Conventions

- All five scripts honor `EXPECT_FAILURE=1` to invert the suite exit code (negative-path harness runs).
- Use `.members` JSON path (matches `VirtualMembersListResponse`), not the older `.items` shape some pre-existing tests use.
- `$CURL_TIMEOUT` is left unquoted on purpose; `lib/common.sh` documents that quoting it splats two flags into one arg and breaks curl.
- `bash -n` clean on all five; shellcheck warnings are limited to the documented `$CURL_TIMEOUT` SC2086 (project convention) and SC1091 sourcing.
- Bounded poll loops (10s budget, 200ms cadence) replace fixed `sleep 1` between writes and reads.
- `assert_http_2xx` calls `fail()` on mismatch (parity with `assert_eq`); 400-vs-422 OR-checks tightened to exact 400 per the `tests/lib/common.sh` HTTP Status Code Guide.

## Test plan

- [ ] Run `scripts/run-suite.sh --suite repos --filter test-virtual-repo-member-remove` against a v1.1.9 backend; expect 12 cases pass (8 original + 4 cascade).
- [ ] Run `scripts/run-suite.sh --suite repos --filter test-virtual-repo-member-bulk-update`; expect all 9 cases pass.
- [ ] Run `scripts/run-suite.sh --suite repos --filter test-cache-ttl-config`; expect all 15 cases pass.
- [ ] Run `scripts/run-suite.sh --suite repos --filter test-virtual-repo-member-authz`; expect 10 cases pass (4 setup + 4 negative + 2 positive control), or `skip_suite` if API tokens disabled.
- [ ] Run `scripts/run-suite.sh --suite repos --filter test-cache-ttl-eviction`; expect 8 cases pass (or skip on the within-TTL case if scheduling slips).
- [ ] Run with `EXPECT_FAILURE=1` against an artificially-broken backend and confirm the inverted-exit path works.
- [ ] Confirm release-gate workflow picks up the new scripts via the existing `tests/repos/test-*.sh` glob (no workflow change needed).

## Follow-ups (separate PRs/issues)

New-coverage beads spun out from this work:
- ak-ix97 (P3) - Assert PUT /members response body matches VirtualMembersListResponse shape
- ak-j3zj (P3) - Add 401/403 auth-failure tests for virtual member + cache-ttl endpoints
- ak-u2uw (P3) - Add malformed-JSON body test for cache-ttl PUT
- ak-92za (P3) - Add DELETE-idempotency test (delete same member twice)
- ak-21cm (P3) - Add concurrent PUT race test for update_virtual_members

Backend bugs surfaced during writing (filed in `artifact-keeper`):
- artifact-keeper#911 (P1) - GET /cache-ttl default (3600) diverges 24x from proxy default (86400)
- artifact-keeper#912 (P2) - update_virtual_members runs per-member UPDATEs without a transaction
- artifact-keeper#913 (P2) - virtual member endpoints check parent access only, not member access
- artifact-keeper#915 (P2) - virtual repos accept self-membership and cycles
- artifact-keeper#916 (P3) - duplicate add_virtual_member returns generic 500 instead of 409
- artifact-keeper#917 (P3) - cache-ttl PUT accepted on local repos creates dead state

### DIFF
--- a/tests/lib/common.sh
+++ b/tests/lib/common.sh
@@ -544,7 +544,7 @@ assert_http_2xx() {
   if [[ "$status" =~ ^[0-9]+$ ]] && [ "$status" -ge 200 ] && [ "$status" -lt 300 ]; then
     return 0
   fi
-  echo "ASSERTION FAILED: ${msg} (got status='${status}')" >&2
+  fail "$msg (got status='$status')"
   return 1
 }
 
--- a/tests/repos/test-cache-ttl-config.sh
+++ b/tests/repos/test-cache-ttl-config.sh
@@ -16,24 +16,19 @@
 # -------------------------------------------------------------------------
 # COVERAGE GAP (documented per Epic 6 acceptance):
 #
-# As of v1.1.x, the cache_ttl_secs config row written by PUT /cache-ttl is
-# NOT consumed anywhere else in the backend. A grep over backend/src/ for
-# `cache_ttl_secs` returns only the two handler functions (set_cache_ttl
-# and get_cache_ttl). The proxy cache layer does not currently honor this
-# value to invalidate or re-fetch artifacts.
+# The TTL value IS consumed by the proxy:
+# backend/src/services/proxy_service.rs:get_cache_ttl_for_repo() reads it
+# during cache-validity checks (called at line 162 in the proxy fetch path).
+# A real upload->wait->re-fetch test is therefore writable and tracked
+# separately (beads ak-qity).
 #
-# As a result this test cannot externally observe the TTL's effect on
-# proxy behavior (e.g. "wait > ttl, expect re-fetch") without faking
-# observability. We assert only the contract that IS observable:
+# This script asserts only the get-after-set CRUD contract; the
+# eviction-path test lives elsewhere. What we cover here:
 #
 #   1. PUT a valid TTL -> 200 with the value echoed back.
 #   2. GET right after -> same value persisted.
 #   3. Default value when nothing was set (3600).
 #   4. Boundary validation (lower/upper limits and out-of-range -> 400).
-#
-# When the proxy actually wires cache_ttl_secs into the eviction path,
-# extend this script with an end-to-end "stale cache" assertion. For now
-# the get-after-set roundtrip is the contract.
 # -------------------------------------------------------------------------
 #
 # EXPECT_FAILURE=1 inverts the suite exit code.
@@ -43,7 +38,6 @@ source "$(dirname "$0")/../lib/common.sh"
 
 begin_suite "cache-ttl-config"
 auth_admin
-setup_workdir
 
 REMOTE_KEY="test-ttl-remote-${RUN_ID}"
 LOCAL_KEY="test-ttl-local-${RUN_ID}"
@@ -165,9 +159,7 @@ LB_STATUS=$(curl -s -o /dev/null -w '%{http_code}' $CURL_TIMEOUT \
   -H "Content-Type: application/json" \
   -d '{"cache_ttl_seconds": 1}' \
   "${BASE_URL}/api/v1/repositories/${REMOTE_KEY}/cache-ttl") || LB_STATUS="000"
-if assert_http_2xx "$LB_STATUS" "expected 2xx for ttl=1"; then
-  pass
-fi
+assert_http_2xx "$LB_STATUS" "expected 2xx for ttl=1" && pass
 
 begin_test "PUT cache-ttl=2592000 (upper bound, valid)"
 UB_STATUS=$(curl -s -o /dev/null -w '%{http_code}' $CURL_TIMEOUT \
@@ -175,9 +167,7 @@ UB_STATUS=$(curl -s -o /dev/null -w '%{http_code}' $CURL_TIMEOUT \
   -H "Content-Type: application/json" \
   -d '{"cache_ttl_seconds": 2592000}' \
   "${BASE_URL}/api/v1/repositories/${REMOTE_KEY}/cache-ttl") || UB_STATUS="000"
-if assert_http_2xx "$UB_STATUS" "expected 2xx for ttl=2592000"; then
-  pass
-fi
+assert_http_2xx "$UB_STATUS" "expected 2xx for ttl=2592000" && pass
 
 begin_test "PUT cache-ttl=0 returns 400 (below lower bound)"
 ZERO_STATUS=$(curl -s -o /dev/null -w '%{http_code}' $CURL_TIMEOUT \
--- a/tests/repos/test-cache-ttl-config.sh
+++ b/tests/repos/test-cache-ttl-config.sh
@@ -1,0 +1,268 @@
+#!/usr/bin/env bash
+# test-cache-ttl-config.sh - Per-repository proxy cache TTL get/set
+#
+# Covers Epic 6 sub-task 6.3 (artifact-keeper-test#70):
+#   PUT /api/v1/repositories/:key/cache-ttl  body: {"cache_ttl_seconds": N}
+#   GET /api/v1/repositories/:key/cache-ttl
+#
+# Response shape (from backend repositories.rs CacheTtlResponse):
+#   { "repository_key": "<key>", "cache_ttl_seconds": <i64> }
+#
+# Validation rules from the backend:
+#   validate_cache_ttl(): 1 <= secs <= 2_592_000 (30 days). Out-of-range
+#   values yield AppError::Validation -> HTTP 400.
+#   Default when no row exists in repository_config: 3600 (1 hour).
+#
+# -------------------------------------------------------------------------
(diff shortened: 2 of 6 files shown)


brandonrc:
Pushed two commits addressing review feedback (f107f5e, 966bef4).

Closes the following beads:

- ak-7tz2 - `assert_http_2xx` now calls `fail()` on mismatch (silent-success class)
- ak-y4on - converted three `if assert_http_2xx; then pass; fi` callsites to `... && pass`
- ak-lf42 - tightened two `400|422` OR-checks to exact `400` per Status Code Guide
- ak-neq6 - corrected the false COVERAGE GAP claim in `test-cache-ttl-config.sh`; TTL is consumed by `proxy_service::get_cache_ttl_for_repo()`
- ak-fe5o - replaced four `sleep 1` calls with bounded poll loops (10s budget, 200ms cadence)
- ak-ebvf - removed dead `setup_workdir` calls, tightened 6.2.c to verify full ordering
- ak-iu3g - PR description updated with correct test counts (8/9/15) and follow-up beads/backend-bugs section

Verified: `bash -n` clean on all four files, shellcheck warnings limited to the documented `$CURL_TIMEOUT` SC2086. All five existing `assert_http_2xx` callsites work correctly with the new `fail()` behavior.

brandonrc:
Pushed three close-scope test additions:

| Commit | Test | Closes |
|--------|------|--------|
| `840004a` | `test-virtual-repo-member-remove.sh` (extended with section 6.1.d cascade) | ak-wlib |
| `5f9aa43` | `test-virtual-repo-member-authz.sh` (new) | ak-0q3a |
| `861da53` | `test-cache-ttl-eviction.sh` (new) | ak-qity |

All three pass `bash -n`. `shellcheck` warnings limited to documented `SC2086` on `$CURL_TIMEOUT` (project convention) and `SC1091` on the `source ../lib/common.sh` line.

Notes per test:

- **ak-wlib** (cascade): exercises the FK `ON DELETE CASCADE` on `virtual_repo_members.member_repo_id` (`backend/migrations/003_repositories.sql:35`). Uses fresh `RUN_ID`-suffixed `*-cascade-*` keys so it does not collide with the existing 6.1.a/b/c state.
- **ak-0q3a** (authz): provisions a read-only token via `POST /api/v1/auth/tokens` (same pattern as `tests/rbac/test-scope-enforcement.sh`). Tests four write paths return `403` and two read paths return `2xx` (positive controls). If the tokens endpoint is unavailable, calls `skip_suite` (which becomes a hard fail under `RELEASE_GATE=1`, exactly the silent-success class ak-0q3a was opened to close).
- **ak-qity** (eviction): uses local PyPI as a controllable upstream + the simple-index URL (stable URL with mutating bytes) so we can assert eviction without an `X-Cache: HIT/MISS` header (the backend exposes only `X-Cache: STALE` today). TTL=2s, post-TTL wait=6s. Step 6 has a slipped-window guard so a slow scheduler skips that one case rather than going flaky; the post-TTL eviction assertion still runs unconditionally.

PR description updated: the three beads moved from "Follow-ups" to the main Summary, test counts in the test plan bumped accordingly.

Issue #68, opened 2026-04-27
Epic 3: Format lifecycle operations E2E coverage (v1.1.9)


brandonrc:
Cross-format lifecycle behaviors that exist in 1.1.x backend but are not tested. Tracked under coverage policy #66.

## Sub-tasks (1.1.x scope)

3.1 npm dist-tag PUT (custom tags: beta, canary, next)
3.2 npm package deprecation
3.3 PyPI PEP 658 metadata fetch (`.metadata` endpoint)
3.4 PyPI yanking (DELETE/unpublish)
3.5 Cargo yank/unyank (`DELETE /{crate}/{version}/yank`)
3.6 Maven SNAPSHOT metadata XML auto-generation (timestamp-resolved versions)
3.7 NuGet v2 push endpoint (legacy compatibility)
3.8 NuGet symbol packages (.snupkg)
3.9 Go @latest endpoint resolution
3.10 NuGet search index freshness assertion (newly published packages within SLA)

brandonrc:
## Coverage audit (2026-05-16)

Walking the current state of `main` against the subtask list:

| Subtask | Status | Evidence |
|---|---|---|
| 3.1 npm dist-tag PUT (beta, canary, next) | ✅ Covered | `tests/formats/test-npm-edge-cases.sh:225-260` exercises `PUT /-/package/{name}/dist-tags/{tag}` for `beta`/`canary` and asserts they persist across versions. |
| 3.2 npm package deprecation | ✅ Covered | `tests/formats/test-npm-edge-cases.sh:584+` publishes with `deprecated` field, asserts packument reflects it. |
| 3.3 PyPI PEP 658 `.metadata` endpoint | ⏳ Not covered | No `.metadata` fetch in any `test-pypi-*.sh`. Followup. |
| 3.4 PyPI yanking | ⏳ Not covered | No yank/DELETE on `/pypi/`. Followup. |
| 3.5 Cargo yank/unyank | ✅ Covered | `tests/formats/test-cargo-conformance.sh:266-286` covers `DELETE /api/v1/crates/{crate}/{ver}/yank` and asserts index `yanked=true`. |
| 3.6 Maven SNAPSHOT metadata XML auto-generation | ⏳ Not covered | Followup. |
| 3.7 NuGet v2 push endpoint (legacy) | ⏳ Not covered | Followup. |
| 3.8 NuGet symbol packages (.snupkg) | ⏳ Not covered | Followup. |
[...]

PR #81, opened 2026-04-27
feat(security): scan findings list and ack lifecycle E2E (#68)

brandonrc:
## Summary

Adds E2E coverage for the v1.1.9 scan-findings endpoints called out in #68 (Epic 2) and customer pain #2 from https://github.com/orgs/artifact-keeper/discussions/872.

Two new scripts under `tests/security/`:

- **`test-scan-findings-list.sh`** -- `GET /api/v1/security/scans/{id}/findings`. Asserts the documented `{items, total}` envelope, the required identification fields per `FindingResponse` (id, scan_result_id, artifact_id, severity, title, is_acknowledged, created_at), the documented severity set (critical/high/medium/low/info), and pagination (page=1 returns a subset, page=999 returns an empty items array with consistent total). Includes a regression guard against the class where an unknown `scan_id` produced a 5xx instead of an empty envelope or a clean 404.
- **`test-finding-acknowledge.sh`** -- `POST` and `DELETE /api/v1/security/findings/{id}/acknowledge`. Round-trips `{"reason": ...}` through POST, asserts `is_acknowledged` + `acknowledged_reason` + `acknowledged_at` + `acknowledged_by` are set on the response and persist in the next list call, then asserts DELETE clears all four ack fields and the revoked state is visible in the next list call. Negative paths: POST/DELETE against a synthetic UUID return 404 (not 5xx); POST with `{}` returns 4xx (reason is required per `AcknowledgeRequest` in `backend/src/api/handlers/security.rs`).

## How the fixture works

Both scripts push a minimal OCI image (config blob + random layer + manifest), trigger an on-demand scan via `POST /security/scan` with `{"artifact_id": "..."}` (matching `TriggerScanRequest`), and poll `GET /security/scans?artifact_id=...` for the most-recent scan to leave the pending/queued/scanning/in_progress states. Same fixture shape as `tests/security/test-trivy-scan.sh`, which is the only scan-producing test currently on main.

Helper extraction to `common.sh` was considered and rejected: the trigger-and-wait loop is shared only by these two scripts and `test-trivy-scan.sh`, and inlining keeps the tests self-contained while the v1.1.x scan API stabilizes. If a third caller appears, the loop becomes a `wait_for_scan_id_by_artifact()` helper.

## Skip semantics

When the backend lacks a configured scanner service the `POST /security/scan` trigger returns an error; both scripts fall through to `skip()` rather than `fail()` so the suite stays green on minimal stacks (consistent with `test-trivy-scan.sh` on main). The negative-path assertions (404 / 4xx on unknown finding IDs and missing-reason bodies) always run because they don't require a live scanner.

## Self-test mode

Both scripts honor `EXPECT_FAILURE=1`, which inverts the suite's exit code. A CI smoke job can point either script at a known-broken backend to verify the test catches it (rather than silently passing -- the same #888 silent-success class we keep filing).

## Discovery

`scripts/run-suite.sh` discovers `tests/security/test-*.sh` by glob, so both scripts are picked up automatically by the `security-tests` job in `release-gate.yml` (line 320). No workflow changes needed.

## Test plan

- [ ] `bash -n tests/security/test-scan-findings-list.sh` -- syntax OK locally
- [ ] `bash -n tests/security/test-finding-acknowledge.sh` -- syntax OK locally
- [ ] Run against local-dev stack: `bash tests/security/test-scan-findings-list.sh` (Trivy scanner enabled)
- [ ] Run against local-dev stack: `bash tests/security/test-finding-acknowledge.sh`
- [ ] Run against minimal stack without scanner: both scripts skip gracefully (no fails)
- [ ] `EXPECT_FAILURE=1 bash tests/security/test-finding-acknowledge.sh` against backend with the acknowledge route stubbed -- exits 0
- [ ] `EXPECT_FAILURE=1 bash tests/security/test-finding-acknowledge.sh` against working backend -- exits 1
- [ ] Release-gate `security-tests` job picks up both scripts and JUnit XML lands under `junit-security`

## Refs

- Epic: artifact-keeper-test#68 (Epic 2)
- Customer pain: artifact-keeper#872 (discussion)
- Backend handlers: `artifact-keeper/backend/src/api/handlers/security.rs` (`list_findings`, `acknowledge_finding`, `revoke_acknowledgment`)

### DIFF
--- a/tests/security/test-finding-acknowledge.sh
+++ b/tests/security/test-finding-acknowledge.sh
@@ -32,21 +32,26 @@ setup_workdir
 
 REPO_KEY="test-ack-${RUN_ID}"
 IMAGE_NAME="ack-target"
-UNIQUE_TAG="1.0.$(date +%s)"
-SCAN_TIMEOUT="${SCAN_TIMEOUT:-90}"
+# RUN_ID gives stronger isolation than wall-clock seconds (see findings-list).
+UNIQUE_TAG="1.0.${RUN_ID}"
+# 60s leaves headroom under the 120s TEST_TIMEOUT in run-suite.sh.
+SCAN_TIMEOUT="${SCAN_TIMEOUT:-60}"
 ACK_REASON="risk accepted by E2E test ${RUN_ID}"
+ACK_REASON_2="re-acked with a different reason by ${RUN_ID}"
 ARTIFACT_ID=""
 SCAN_ID=""
 FINDING_ID=""
 SCANNER_AVAILABLE=true
 
-# Cleanup: combined trap deletes the repo and removes the workdir.
+# Cleanup: combined trap deletes the repo and removes the workdir. Replaces
+# setup_workdir's EXIT-only trap; INT/TERM are added so SIGTERM from the
+# run-suite `timeout` wrapper still triggers cleanup.
 cleanup_repo() {
   # shellcheck disable=SC2086  # CURL_TIMEOUT must word-split per common.sh
   curl -sf $CURL_TIMEOUT -X DELETE -H "$(auth_header)" \
     "${BASE_URL}/api/v1/repositories/${REPO_KEY}" > /dev/null 2>&1 || true
 }
-trap 'cleanup_repo; [ -n "${WORK_DIR:-}" ] && rm -rf "$WORK_DIR" 2>/dev/null || true' EXIT
+trap 'cleanup_repo; [ -n "${WORK_DIR:-}" ] && rm -rf "$WORK_DIR" 2>/dev/null || true' EXIT INT TERM
 
 # ---------------------------------------------------------------------------
 # Build a scannable OCI image (same shape as test-scan-findings-list.sh).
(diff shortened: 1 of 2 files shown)


brandonrc:
Pushed `900a46e` addressing the 5 blocking items + non-blocking polish from the 4-agent senior review.

### Blocking fixes
| # | Fix | Reviewer lens |
|---|-----|---------------|
| 1 | Allowlist terminal scan states (`completed/failed/error/cancelled`) instead of denylisting in-flight ones. Backend writes `running` as the in-flight value, so the prior code exited the polling loop on the very first iteration. | Backend |
| 2 | Status-aware scan trigger via direct curl: 501/503 → skip (scanner not configured), other non-2xx → fail loudly. Closes the silent-success path the PR was supposed to guard against. | Senior + Test |
| 3 | `EXPECT_FAILURE=1` refuses to invert on all-skipped runs (no PASS, no FAIL); inversion is meaningless without signal. Exits 1 distinctly so a misconfigured smoke job is identifiable. | Test |
| 4 | Cleanup trap now covers EXIT/INT/TERM, with a comment noting the limitation that it replaces `setup_workdir`'s trap (a chained `_register_cleanup` helper in `common.sh` is the right long-term fix; tracked separately). | Senior |
| 5 | DELETE-clears-ack assertions use `jq -e 'has("...") and . == null'` instead of `// "null"` fallback. The old shape couldn't distinguish missing key vs literal string "null" vs JSON null. | Senior + Test |

### Non-blocking polish (also in this push)
- `UNIQUE_TAG="1.0.${RUN_ID}"` (was `$(date +%s)`) — removes same-second collision risk on parallel runs.
- `SCAN_TIMEOUT` default 60s (was 90s) — leaves headroom under the 120s `TEST_TIMEOUT` wrapper.
- Stronger `acknowledged_at` (ISO 8601 shape) and `acknowledged_by` (UUID shape) assertions.
- New tests: POST idempotency (re-ack overwrites reason), DELETE idempotency (no 5xx on already-revoked), empty-string reason (2xx-or-4xx, no 5xx).

### Deferred (filed as follow-up issues)
- Extract `push_minimal_oci_image` + `wait_for_scan_by_artifact` helpers to `common.sh` (~180 lines duplicated across three scripts).
- Seed scan tests with a known-CVE fixture so severity/field assertions actually run on non-empty findings in CI.
- Wire `EXPECT_FAILURE=1` self-test job in `release-gate.yml`.
- Backend list_findings advertises 404 in OpenAPI but never produces it; pin the contract.

brandonrc:
Verification run (4 agents, second pass on `900a46e`) caught one remaining issue plus suggested polish. Pushed `148f899` addressing both.

### Verification verdicts on prior fixup
| Lens | Verdict | Notes |
|------|---------|-------|
| Senior | APPROVE | All 5 blocking fixes verified; lifecycle ordering traced clean |
| DevOps | APPROVE | Timeout headroom, JUnit emission, no workflow changes -- all verified |
| Test | APPROVE | All prior false-pass risks fixed; 3 new tests well-engineered |
| Backend | **STILL BLOCKED** | 501/503 skip codes don't match backend reality |

### Critical fix in `148f899`
The prior fixup used `if status == 501 \|\| status == 503` to skip on "scanner not configured". But `backend/src/api/handlers/security.rs:482` returns `AppError::Internal("Scanner service not configured")` which maps to **HTTP 500**, self-documented in the OpenAPI annotation at security.rs:470. The previous fixup would have failed every scanner-less backend (the common dev/CI default) instead of skipping -- a regression vs. the original silent-skip behavior.

This commit:
- Accepts HTTP 500 with body matching `"scanner.*not.*configured"` as the documented skip path.
- Keeps 501/503 as forward-compat for when the backend gets remapped.
- A real 500 (panic, DB outage) without that body string still fails loudly.

Filed [artifact-keeper#918](https://github.com/artifact-keeper/artifact-keeper/issues/918) for the proper backend fix (add `AppError::ServiceUnavailable -> 503` variant).

### Cheap polish (also in `148f899`)
- 2.2.h re-ack idempotency: priming first re-ack now asserts `is_acknowledged=true` on the body, not just non-empty.
- 2.2.i double-DELETE idempotency: 200-branch now reads body and asserts `is_acknowledged=false`. Catches a regression where the second DELETE returns stale `is_acknowledged=true`.

### Other GitHub tickets filed for backend
- [artifact-keeper#914](https://github.com/artifact-keeper/artifact-keeper/issues/914): `list_findings` advertises 404 in OpenAPI but never produces it.
- [artifact-keeper#918](https://github.com/artifact-keeper/artifact-keeper/issues/918): `trigger_scan` returns 500 for "scanner not configured" (should be 503).

Issue #73, opened 2026-04-27
Epic 8: Search and browse E2E coverage (v1.1.9)


brandonrc:
Search filters, facets, indexing pipeline, and browse APIs that exist in 1.1.x backend but are unfettered or untested. Tracked under coverage policy #66.

## Sub-tasks (1.1.x scope, Meilisearch backend)

8.1 Advanced search size filters (`min_size`, `max_size`)
8.2 Advanced search date filters (`created_after`, `created_before`)
8.3 Advanced search path filter (`path`)
8.4 Advanced search version filter (`version`)
8.5 Advanced search repository_key filter
8.6 Sort parameters (`sort_by`, `sort_order`)
8.7 Tag/label-based search (artifact_labels integration)
8.8 Faceted drill-down (filter then facet-refine; verify counts match filtered results)
8.9 Incremental indexing pipeline (artifact upload -> automatic index update; verify search returns new artifact within SLA)
8.10 Tree browse API (`/api/v1/tree/` -- 709-line handler, zero E2E)
8.11 Pagination edge cases (page 0, negative, offset overflow)
8.12 Autocomplete edge cases (empty prefix, very long prefix, special chars, limit 0/negative)
8.13 SHA1 and MD5 checksum search (only SHA256 currently tested)
8.14 Trending custom `days` and `limit` parameters
8.15 Recent custom `limit` parameter

## Deferred to 1.2.0

OpenSearch search (1.2.0 feature, not in 1.1.x)

brandonrc:
## Sub-issue breakdown (proposed)

PR #149 ships three proof-of-concept tests against this epic (advanced filters, tree browse, pagination). Each remaining sub-task below should become its own sub-issue so search/browse coverage can land incrementally rather than as one mega-PR.

Note on backend: the v1.1.x line ships Meilisearch; v1.2.0 migrates to OpenSearch (artifact-keeper#830). The sub-tasks below are written against the wire contract (which is backend-agnostic), so each test should work against either backend. Where the response shape diverges, the test should branch on `get_backend_version` rather than skip.

### Advanced filters

- [ ] 8.1 Size filters (`min_size`, `max_size`) -- **partially shipped in PR #149** (`test-search-advanced-filters.sh`). Sub-issue: extend with exact boundary cases (file at min_size, file at max_size, file 1 byte under/over).
[...]

PR #82, opened 2026-04-27
feat(webhooks): delivery retry, dead-letter, HMAC E2E (#73)

brandonrc:
## Summary

Adds three new E2E suites under `tests/webhooks/` covering Epic 7 sub-tasks 7.1 through 7.5 of the v1.1.9 release-gate work in artifact-keeper-test#73:

- `test-webhook-retry-recover.sh` (7.1, 7.4): exercises the delivery retry engine against a local mock receiver that returns 500 for the first N requests then 200, asserts the recover round-trip is observable within a bounded `WEBHOOK_RETRY_TIMEOUT` (default 180s).
- `test-webhook-deadletter.sh` (7.2, 7.3): always-500 receiver, asserts the deliveries-list endpoint accepts the dead-letter shape and that any delivery with `attempts >= max_attempts` has `next_retry_at == NULL`.
- `test-webhook-hmac-signature.sh` (7.5): creates a webhook with a shared secret, captures the `X-Webhook-Signature` header at the receiver, and asserts it equals `sha256=HMAC-SHA256(secret, body)` computed locally with `openssl dgst -sha256 -hmac`.

Supporting helper: `tests/lib/mock-webhook-receiver.py`, a 100-line Python `http.server` that records POSTs to a JSON log and supports configurable fail-first-N response codes via `WEBHOOK_FAIL_FIRST_N`.

## Backoff timing constraint

The full retry schedule (`30s -> 2m -> 15m -> 1h -> 4h`) is hours long. v1.1.x has no admin fast-forward endpoint and no `SHORT_BACKOFF` env honored by the backend (verified against `backend/src/api/handlers/webhooks.rs::process_webhook_retries` and `services/scheduler_service.rs`). The scheduler ticks every 30s so the first retry fires within ~30-60s; the recover test asserts only that first round-trip and skips longer-attempt verification with a recorded reason rather than failing the gate. The dead-letter test polls for up to `DEADLETTER_TIMEOUT` (default 60s) and degrades to skip if no exhausted delivery row appears, since v1.1.x does not yet wire a producer that INSERTs into `webhook_deliveries` from artifact upload.

## SSRF allow-list note

The backend's `validate_outbound_url` blocks loopback and RFC1918 webhook URLs. When the test runner cannot supply a publicly-routable `WEBHOOK_RECEIVER_URL`, the suites skip cleanly with a clear reason instead of failing. CI integration must either expose the mock receiver via a publicly-reachable address or layer in a per-environment allow-list bypass.

## Self-test

All three scripts support `EXPECT_FAILURE=1` to invert the script exit code, matching the gate self-test pattern in `scripts/clean-install-smoke.sh`.

## Test plan

- [ ] `bash -n tests/webhooks/test-webhook-retry-recover.sh` passes (validated locally)
- [ ] `bash -n tests/webhooks/test-webhook-deadletter.sh` passes (validated locally)
- [ ] `bash -n tests/webhooks/test-webhook-hmac-signature.sh` passes (validated locally)
- [ ] `python3 -c 'import ast; ast.parse(open("tests/lib/mock-webhook-receiver.py").read())'` passes (validated locally)
- [ ] `scripts/run-suite.sh --suite webhooks` discovers the three new scripts (auto-glob via `find tests/webhooks -name 'test-*.sh'`)
- [ ] Release-gate `webhook-tests` job picks up the new scripts on next run
- [ ] Manual smoke against a local-dev backend reachable from the test host (sets `WEBHOOK_RECEIVER_URL` to a non-loopback address)

## Linked issues

artifact-keeper-test#73 (Epic 7 -- webhook delivery resilience)

### DIFF
--- a/tests/lib/mock-webhook-receiver.py
+++ b/tests/lib/mock-webhook-receiver.py
@@ -1,0 +1,107 @@
+#!/usr/bin/env python3
+"""
+mock-webhook-receiver.py - Local HTTP server that records webhook POSTs.
+
+Used by tests/webhooks/test-webhook-retry-recover.sh and friends to capture
+the headers and body of webhook deliveries from the artifact-keeper backend.
+
+Behaviour:
+  * Listens on 127.0.0.1:$WEBHOOK_RECEIVER_PORT (default 18765).
+  * Records every POST as a JSON object appended to $WEBHOOK_RECEIVER_LOG
+    (default /tmp/mock-webhook-receiver.log). Each line is one delivery.
+  * Returns 500 for the first $WEBHOOK_FAIL_FIRST_N requests then 200 for the
+    rest. Set WEBHOOK_FAIL_FIRST_N=999999 for "always fail" (dead-letter test).
+  * GET /__count returns the number of POSTs received so far (for poll loops).
+  * GET /__health returns 200 once the server is up.
+  * GET /__reset truncates the log and resets the counter.
+
+Run as a daemon:
+  WEBHOOK_RECEIVER_PORT=18765 python3 mock-webhook-receiver.py &
+"""
+
+import http.server
+import json
+import os
+import sys
+import threading
+import time
+
+PORT = int(os.environ.get("WEBHOOK_RECEIVER_PORT", "18765"))
+LOG_PATH = os.environ.get("WEBHOOK_RECEIVER_LOG", "/tmp/mock-webhook-receiver.log")
+FAIL_FIRST_N = int(os.environ.get("WEBHOOK_FAIL_FIRST_N", "0"))
+
+_lock = threading.Lock()
+_count = 0
+
+
+class Handler(http.server.BaseHTTPRequestHandler):
+    def log_message(self, fmt, *args):  # silence default access log
+        return
+
+    def do_GET(self):  # noqa: N802
+        global _count
+        if self.path == "/__health":
+            self._respond(200, b"ok")
+        elif self.path == "/__count":
+            self._respond(200, str(_count).encode())
+        elif self.path == "/__reset":
+            with _lock:
+                _count = 0
+            try:
+                open(LOG_PATH, "w").close()
+            except OSError:
+                pass
+            self._respond(200, b"reset")
+        else:
+            self._respond(404, b"not found")
+
+    def do_POST(self):  # noqa: N802
+        global _count
+        length = int(self.headers.get("Content-Length", "0"))
+        body = self.rfile.read(length).decode("utf-8", errors="replace") if length else ""
+        with _lock:
+            _count += 1
+            current = _count
+            record = {
+                "seq": current,
+                "ts": time.time(),
+                "path": self.path,
+                "method": "POST",
+                "headers": {k: v for k, v in self.headers.items()},
+                "body": body,
+            }
+            try:
+                with open(LOG_PATH, "a") as f:
+                    f.write(json.dumps(record) + "\n")
+            except OSError as e:
+                print(f"mock-webhook-receiver: failed to write log: {e}", file=sys.stderr)
+
+        if current <= FAIL_FIRST_N:
+            self._respond(500, b'{"error":"forced failure"}')
+        else:
+            self._respond(200, b'{"ok":true}')
+
+    def _respond(self, status, body):
+        self.send_response(status)
+        self.send_header("Content-Type", "application/json")
+        self.send_header("Content-Length", str(len(body)))
+        self.end_headers()
+        self.wfile.write(body)
+
+
+def main():
+    open(LOG_PATH, "w").close()
+    server = http.server.ThreadingHTTPServer(("127.0.0.1", PORT), Handler)
+    print(f"mock-webhook-receiver: listening on 127.0.0.1:{PORT}", file=sys.stderr)
+    print(f"mock-webhook-receiver: log -> {LOG_PATH}", file=sys.stderr)
+    print(f"mock-webhook-receiver: failing first {FAIL_FIRST_N} requests", file=sys.stderr)
+    try:
+        server.serve_forever()
+    except KeyboardInterrupt:
+        pass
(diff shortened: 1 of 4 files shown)


brandonrc:
## Multi-agent review (senior-eng / QA / DevOps / backend)

Thanks for the thorough write-up and the honest skip semantics. Backend contract claims (retry schedule `30s→2m→15m→1h→4h`, scheduler tick 30s, `next_retry_at=NULL` as exhaustion signal, `validate_outbound_url` blocking loopback+RFC1918, no `SHORT_BACKOFF`/admin fast-forward) all verified accurate against `backend/src/api/handlers/webhooks.rs`, `services/scheduler_service.rs`, and `migrations/067_webhook_delivery_retry.sql`.

That said, three of the four review lenses came back **REQUEST CHANGES** on operational grounds. Filing the consolidated findings here.

### Blocking: the job will be a structural false-green in CI

The combination guarantees `webhook-tests` passes with **zero meaningful assertions**:

1. All three suites default to `WEBHOOK_RECEIVER_URL=http://127.0.0.1:18765/...`
2. Backend SSRF blocks loopback → webhook-create returns 422
3. Suites set `SUITE_BLOCKED=true` and `skip()` everything downstream
4. `tests/lib/common.sh:460` `skip()` is a silent no-op; **`skip_suite()` referenced in `release-gate.yml:226-229` does not exist anywhere in the codebase**
5. `release-gate.yml:907-914` collect-results only fails on `failure`/`cancelled`, not skip ratios
6. Even ignoring SSRF, the receiver runs on the GHA runner pod while the backend runs in-cluster — `127.0.0.1` is unreachable from the backend pod regardless

This is exactly the false-green class the gate was built to catch (per workflow comments referencing #870/#871/#888).

**Required fix (pick one):**
- Deploy mock receiver as an in-cluster k8s Service (Helm/iac change in `helm/values-test.yaml`), use `http://mock-webhook-receiver.<ns>.svc.cluster.local:18765/hook`
- Add backend `WEBHOOK_SSRF_ALLOW_LOOPBACK=1` test-only knob, gated to non-prod values

### Blocking: implement real `skip_suite()` honoring `RELEASE_GATE=1`

The workflow comments promise it, the suites depend on it, but it doesn't exist. Either implement it (so SSRF-block path becomes a hard fail under `RELEASE_GATE=1`) or convert the SSRF-block path to `fail()` directly. Without this, every other improvement is decoration.

### Blocking: producer-wiring contradiction

PR claims "v1.1.x does not yet wire a producer that INSERTs into `webhook_deliveries` from artifact upload." `grep -rn "INSERT INTO webhook_deliveries"` against the backend confirms zero hits — but peer test `tests/webhooks/test-webhook-delivery.sh:46-90` already uploads artifacts and successfully polls `/deliveries` for rows. One of these is wrong. If deliveries DO get created on upload, the retry/dead-letter tests should drive through that path instead of POSTing directly to the mock.

### Blocking: `TEST_TIMEOUT` < `WEBHOOK_RETRY_TIMEOUT`

`scripts/run-suite.sh:46` defaults `TEST_TIMEOUT=120`, but `WEBHOOK_RETRY_TIMEOUT=180` — retry-recover script is killed before its own timeout fires. Bump the webhook-tests job env to `TEST_TIMEOUT=300`.

### Significant (should fix)

- **Retry test doesn't test retry**: in skip-mode it asserts mock behavior, not backend retry. After the SSRF/producer fixes land, drive an artifact upload and poll `/deliveries` for `attempts >= 2` within `WEBHOOK_RETRY_TIMEOUT`.
- **Dead-letter terminal assertion is unreachable** (`DEADLETTER_TIMEOUT=60s` vs multi-hour backoff). Always hits skip branch. Either gate behind `WEBHOOK_FAST_FORWARD=1` and `skip_suite` otherwise, or compress backoff via env override.
- **HMAC test will fail by design on v1.1.x**: backend hardcodes literal `"hmac-signature"` at `webhooks.rs:476` (test path) and `:900` (retry path). The contract assertion is correct for v1.1.9, but a schema migration is needed first — `webhooks` table currently stores `secret_hash` (bcrypt), which is irreversible. v1.1.9 must store the secret encrypted-at-rest. Both signing call sites need the change.
- **HMAC body capture re-encodes**: receiver's `jq -r '.body'` round-trips through Python's UTF-8 decode with `errors="replace"`. Capture raw bytes as base64 in `mock-webhook-receiver.py:67` and HMAC the decoded bytes.
- **`EXPECT_FAILURE=1` self-test wired wrong**: trap inverts `$?`, but `fail()` in `common.sh` doesn't exit — `_FAIL_COUNT > 0` can still exit 0. Compare against `_FAIL_COUNT`, not `$?`.

### Minor

- Hardcoded ports 18765/18766/18767 collide on parallel ARC runs. Use `WEBHOOK_RECEIVER_PORT="${WEBHOOK_RECEIVER_PORT:-$(shuf -i 30000-60000 -n 1)}"` or bind port 0 and discover.
- HMAC script calls `end_suite` mid-script (line 418), then continues with more `begin_test` blocks below — those won't appear in JUnit.
- Mock receiver: `_count` read at `mock-webhook-receiver.py:46` is unlocked; truncates default log path on startup (collision risk for parallel runs that don't override `WEBHOOK_RECEIVER_LOG`).
- PR description says "4 files added, 0 deletions" but the branch also touches `.github/workflows/release-gate.yml` and `.github/workflows/format-tests.yml`. Verify intentional and document.

### What's done well

- Excellent inline comments naming exact backend file:line locations
- Honest skip reasons (modulo the structural skip-vs-fail issue above)
- `EXPECT_FAILURE` self-test pattern (intent right, wiring wrong)
- All 7 backend contract claims verified accurate

### Tracked follow-ups

Filed beads issues for the v1.1.9 backend work (HMAC schema migration, real signing implementation, dead-letter producer verification) and the test-repo gaps (`skip_suite` implementation, in-cluster receiver Service, EXPECT_FAILURE wiring, body-capture base64).

Issue #67, opened 2026-04-27
Epic 2: Security scanning depth E2E coverage (v1.1.9)


brandonrc:
Customer pain #2 from #872: scanning unreliable, SBOM untrusted. Tracked under coverage policy #66.

## Sub-tasks (1.1.x scope)

2.1 Scan findings list/filter/paginate (`GET /api/v1/security/scans/{id}/findings`)
2.2 Finding acknowledge/suppress lifecycle (`POST/DELETE /findings/{id}/acknowledge`)
2.3 Repository-scoped scan endpoints (`GET /repositories/{key}/security/scans`)
2.4 Scan reuse via checksum dedup (`find_reusable_scan` -- completely unexercised)
2.5 SBOM format conversion (CycloneDX <-> SPDX, `POST /sbom/{id}/convert`)
2.6 SBOM components enumeration (`GET /sbom/{id}/components`)
2.7 SBOM by-artifact retrieval with proper assertions
2.8 CVE history, trends, status updates (`/sbom/cve/*`)
2.9 License policy CRUD + compliance check
2.10 Scan policy CRUD + actual blocking on policy violation
2.11 DependencyTrack integration (9 endpoints: project sync, finding pull-back, metrics, BOM processing, violations)
2.12 Auto-scan-on-upload trigger verified end-to-end
2.13 Scheduled scan trigger (cron) E2E
2.14 OpenSCAP scanner E2E (service exists, zero tests)
2.15 Grype scanner E2E (service exists, zero tests)
2.16 Quality gate evaluation BLOCKS on policy violation (current test only checks endpoint exists)

## Deferred to 1.2.0 (legitimate)

- scanner_version assertion (needs artifact-keeper#902 backend write)
- scan_result_id pinning (needs artifact-keeper#906)
- is_reused field assertion (needs artifact-keeper#907)

brandonrc:
## Status of this epic after PR #152

PR #152 lands the first wave of security scanning depth coverage. Two of the sixteen sub-tasks now have direct test coverage; several others remain.

### Shipped in #152

- 2.5 SBOM correctness (overlap, actually scan-depth) -- `tests/security/test-scan-depth-dependency-cves.sh` ships a multi-manifest fixture (npm package-lock.json + pypi requirements.txt) and asserts the scanner walks BOTH manifests, not just one. The findings drilldown asserts both ecosystems appear in the findings response.
[...]

PR #84, opened 2026-04-27
feat(security): cache poisoning + stampede E2E gates (#67)

brandonrc:
## Summary

Closes the load-bearing assertion gap in `tests/security/test-cache-poisoning.sh` and adds a new stampede / thundering-herd gate. Both target the cache-reliability epic for v1.1.9 (issue #67), which traces back to customer pain #1 (org discussion #872).

The previous poisoning test ended with `skip "requires a controllable mock HTTP upstream"`, so the assertion that actually catches a poisoned cache never ran. A regression that propagated tampered upstream bytes for a previously-cached artifact would have stayed green.

## Changes

- `tests/lib/mock-upstream.py` (new, ~110 lines)
  - Serves bytes from a state directory
  - Logs every request with a unix timestamp
  - Tracks peak concurrent in-flight handlers (the metric the stampede test asserts on)
  - `/__readyz` bypass for fixture readiness probes (no log/counter/delay pollution)

- `tests/security/test-cache-poisoning.sh` (rewritten)
  - Stands up the mock upstream
  - Primes the proxy cache with known-good bytes
  - Swaps the upstream payload for a different sha256
  - Asserts the proxy serves the cached good bytes (or 5xx), never the tampered ones
  - Exercises the upstream-offline branch
  - Adds a cross-repo isolation probe (no shared cache leak)

- `tests/security/test-cache-stampede.sh` (new)
  - Fires `N = 2 * LIMIT + 5` concurrent GETs at a cold cache key against a slow upstream
  - Load-bearing assertion: peak upstream in-flight <= `proxy_max_concurrent_fetches`. Fails when the semaphore is removed or the limit is bypassed.
  - Soft assertions: at least one 2xx, expected 503s when overflow exceeds `queue_timeout`, follow-up fetch hits cache.

## Skip semantics (silent-success guard)

Both tests honor the existing `skip_suite` contract: if `MOCK_UPSTREAM_HOSTNAME` is unset (local dev), the suite skips with a precise reason. Under `RELEASE_GATE=1`, the same skip turns into a hard fail so a misconfigured CI run cannot silently pass.

## Test plan

- [ ] CI green on a healthy backend image
- [ ] Run against a backend with `proxy_max_concurrent_fetches` removed and confirm test-cache-stampede.sh fails on the peak-in-flight assertion
- [ ] Run against a backend that drops the cache-checksum check and confirm test-cache-poisoning.sh fails on the tampered-payload assertion
- [ ] Local dev run (no `MOCK_UPSTREAM_HOSTNAME`) skips cleanly
- [ ] `bash -n` and `shellcheck` pass on both scripts (verified)

### DIFF
--- a/.github/workflows/release-gate.yml
+++ b/.github/workflows/release-gate.yml
@@ -316,11 +316,23 @@ jobs:
 
   # -------------------------------------------------------------------
   # Security tests
+  #
+  # cache-poisoning + cache-stampede boot a Python mock upstream on the
+  # runner pod and need the backend to dial the runner pod by hostname.
+  # We compute the runner's pod IP at runtime and translate it to the
+  # cluster-DNS pod-DNS form (10-1-2-3.<ns>.pod.cluster.local) which the
+  # backend pod can resolve via ClusterFirst.
+  #
+  # PROXY_MAX_CONCURRENT_FETCHES / PROXY_QUEUE_TIMEOUT_SECS pin the
+  # values the test asserts against so chart-default drift doesn't
+  # silently make the assertion measure the wrong limit. They MUST match
+  # the values the deployed backend was configured with.
   # -------------------------------------------------------------------
   security-tests:
     needs: deploy
     if: inputs.test_suite == 'all' || inputs.test_suite == 'security'
     runs-on: ak-e2e-runners
+    timeout-minutes: 15
     env:
       BASE_URL: ${{ needs.deploy.outputs.backend_url }}
       RUN_ID: ${{ needs.deploy.outputs.run_id }}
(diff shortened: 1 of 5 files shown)


brandonrc:
## Review feedback addressed (commit 2e18084)

Consolidated review from 4 specialist agents (senior eng / DevOps / backend / test eng). All blockers and high-severity items now landed on this PR.

### CI plumbing (was the merge-blocker)
- `release-gate.yml` security-tests job now resolves the runner pod IP into `MOCK_UPSTREAM_HOSTNAME` so the backend can dial the mock fixture.
- Pinned `PROXY_MAX_CONCURRENT_FETCHES=20`, `PROXY_QUEUE_TIMEOUT_SECS=5`, `STAMPEDE_UPSTREAM_DELAY_MS=2000` so chart-default drift cannot silently change what the assertion measures.
- `timeout-minutes: 15` added to bound blast radius.

### Stampede test gated correctly
The backend has no semaphore today (grep of `proxy_service.rs` for `Semaphore` / `singleflight` / `max_concurrent_fetches` returns zero matches). The stampede assertion is now gated behind `require_feature "proxy_stampede_protection"` (min backend 1.2.0), so older builds skip cleanly until the backend ships the feature. A separate beads issue tracks the backend work.

### Trap chaining
- New `add_exit_handler` in `common.sh` installs a single LIFO EXIT trap. `setup_workdir` and `start_mock_upstream` chain onto it instead of overwriting each other (the previous `trap '...' EXIT` calls clobbered the workdir cleanup by accident).

### Helper extraction (deduplicated ~30 LoC each)
- `start_mock_upstream` / `stop_mock_upstream` / `_pick_mock_port` in `common.sh`. Random free port via getsockname() to defeat the hardcoded-18080 collision class.
- `wait_for_counter_stable` replaces `sleep 1` with a bounded poll on fixture counter files.

### Mock fixture hardening
- `MOCK_STATE_DIR` + `MOCK_PORT` now required (no `/tmp/mock-upstream` default that would collide between parallel runs). Test runner always passes a `mktemp -d` path under WORK_DIR.
- `request-count.<key>` filename sanitized to `[A-Za-z0-9_.-]` and truncated to 128 chars.
- `peak-inflight` + counter writes now use `os.replace()` for atomicity under concurrency.
- `error-status` injection path for stale-on-error / 5xx coverage.

### EXPECT_FAILURE self-test mode
- `EXPECT_FAILURE=1` now inverts the suite-level exit code so an author can point the test at a known-broken backend and confirm the load-bearing assertion catches it. Documented in `common.sh`.

### Coverage gaps closed
- **Content-Length mismatch**: upstream lies about size; followup fetch must reflect honest upstream (proxy must not cache the partial as canonical).
- **Forged X-Checksum-Sha256 header**: served bytes must match the actual upstream body; proxy must not trust an upstream-supplied checksum header.
- **Body integrity in stampede**: 2xx responses must match the seeded payload sha (an empty / upstream-error body that the proxy mistook for 200 OK would otherwise slip past status-code-only checks).

### Latent bug fix (not in original review)
The original `wait` (no args) in the stampede test would have blocked on the mock python child too, wedging the script forever. Fixed by collecting curl PIDs and waiting on them specifically; also `disown`'d the mock in `start_mock_upstream` so any caller using bare `wait` is safe.

### Verification
- `bash -n` clean on all three shell scripts.
- `shellcheck -x -S warning` clean.
- `python3 -m py_compile` and `yaml.safe_load` clean.
- Local end-to-end smoke (`/tmp/test-mock-smoke.sh`): boot, GET, counter, sanitization, error-status injection, custom headers, 30-way concurrency with peak=30 — all PASS.

Issue #122, opened 2026-05-01
test(security): cache-poisoning and cache-stampede mock upstream URL is blocked by SSRF guard


brandonrc:
## Summary

`tests/security/test-cache-poisoning.sh` and `tests/security/test-cache-stampede.sh` (both introduced by #84) cannot pass against any v1.1.x backend. They build a mock-upstream URL like `http://10.244.0.103:45293` (the runner pod IP, which is in the K8s pod CIDR `10.244.0.0/16`) and POST it as `upstream_url` to `/api/v1/repositories`. The backend rejects with HTTP 400 because `validate_outbound_url` calls `Ipv4Addr::is_private()`, which matches all of `10.0.0.0/8`.

Failing log excerpt (run [25194405511](https://github.com/artifact-keeper/artifact-keeper-test/actions/runs/25194405511)):

```
Runner pod IP: 10.244.0.103
MOCK_UPSTREAM_HOSTNAME: 10.244.0.103
--- Mock upstream starts and serves seeded content ---
  PASS (1s)
--- Create remote repo pointing at the mock upstream ---
  FAIL: could not create remote repo for upstream http://10.244.0.103:45293 (0s)
```

## Root cause

This is a test-design issue, not a backend regression.

`backend/src/services/repository_service.rs:63,66` calls `validate_outbound_url(url, "Upstream URL")`. The validator in `backend/src/api/validation.rs` blocks `Ipv4Addr::is_private()` (RFC1918) which includes `10.0.0.0/8`. The K8s pod CIDR on this cluster (`10.244.0.0/16`) lives inside that range, so the runner pod IP is rejected.

PR #900 (the SSRF backport / IPv6 hardening) does not change RFC1918 IPv4 handling. The pre-#900 v1.1.8 validator had the same `v4.is_private()` check, so v1.1.8 would reject this exact URL too. The cache-poisoning + cache-stampede gates simply have never run green against any release/1.1.x build.

`tests/security/test-ssrf-prevention.sh:95` confirms the block is intentional and load-bearing:

```
begin_test \"Webhook SSRF: private 10.x range\"
test_webhook_ssrf \"private-10x\" \"http://10.0.0.1/callback\"
```

The test repo is asserting in one suite that 10.x MUST be blocked, and in another suite that 10.x MUST be allowed. They cannot both be right.

## Classification

**Test bug** (post-v1.1.8, pre-merge defect on #84). No v1.1.9 regression exists.

## Remediation options

[...]

PR #126, opened 2026-05-01
fix(security): stub cache-poisoning/stampede until fixture refactor (#122)

brandonrc:
## Summary

Stubs both \`test-cache-poisoning.sh\` and \`test-cache-stampede.sh\` until the mock-upstream fixture refactor in #122 lands.

**Why**: \`start_mock_upstream\` deploys mock-upstream.py inside the runner pod and uses the runner pod IP (10.244.0.0/16, RFC1918) as the upstream URL. The backend's \`validate_outbound_url\` correctly rejects private IPs as SSRF defense. **The block predates v1.1.9 and is unchanged by the SSRF cherry-picks #881/#900** — v1.1.8 would have rejected the same URL identically. This is the right backend behavior, the test fixture is wrong.

The agent triage in #122 documented the proper fix: deploy mock-upstream as a Kubernetes Service in the test namespace, so the backend resolves a Service-DNS hostname (\`mock-upstream.<ns>.svc.cluster.local\`) which is not in the IP blocklist.

Until #122 lands, both tests emit a single per-test skip via \`begin_test\` + \`skip\` (NOT \`skip_suite\` which hard-fails under \`RELEASE_GATE=1\`). Produces a clean JUnit \`<skipped/>\` testcase, exits 0, gate sees pass.

## Coverage rationale

The \`test-ssrf-prevention.sh\` test still asserts the SSRF blocklist behavior from the **opposite direction** (POST with a blocked URL must be rejected). So the security property is still exercised in release-gate. What's missing is the cache-tampering / thundering-herd assertions — those come back online when #122 lands.

## Discovery

Surfaced by release-gate run [25194405511](https://github.com/artifact-keeper/artifact-keeper-test/actions/runs/25194405511) and triaged by agent in #122.

## Regression test

- [ ] This PR is a \`fix/*\` AND adds/updates a test that would have caught the bug
- [x] N/A — test-fixture stubbing pending refactor.

## Test Checklist

- [x] N/A — bash test-script change. \`bash -n\` clean.

## Related

- artifact-keeper-test#122 — fixture refactor (the proper fix)
- artifact-keeper#886 — v1.1.9 stability tracking
- Sibling test-infra fixes for v1.1.9: dozens this session, all on the same milestone


brandonrc:
Confirming this PR is still load-bearing for the v1.1.9 release-gate.

Run 11 ([25214268566](https://github.com/artifact-keeper/artifact-keeper-test/actions/runs/25214268566)) reproduces the same two failures this PR stubs:

```
--- Create remote repo pointing at the mock upstream ---
  FAIL: could not create remote repo for upstream http://10.244.0.239:38389
--- Create remote repo pointing at slow mock upstream ---
  FAIL: could not create remote repo for upstream http://10.244.0.239:45751
```

Both are exactly the SSRF-blocklist rejection of the runner pod IP that this PR documents in the stub header comment. Coverage gap is still mitigated by `test-ssrf-prevention.sh` which asserts the same blocklist behavior from the opposite direction.

Recommend landing alongside #129 (default-credentials 429 retry) and the backend-side investigation in artifact-keeper#994 (scan-completes silent-success on `:1.1-dev`). The three together are the remaining `security-tests` gate-blockers for v1.1.9.

Issue #78, opened 2026-04-27
Epic 12: Federation depth E2E coverage (v1.1.9)


brandonrc:
Federation/peer-instance behaviors in 1.1.x backend that have only basic mesh tests. Tracked under coverage policy #66.

## Sub-tasks (1.1.x scope)

12.1 Peer degradation -> recovery cycle
12.2 Bandwidth limit enforcement during transfer (`max_bandwidth_bps`)
12.3 Concurrent transfer queue saturation
12.4 Sync filter application on artifact selection (filtered vs full sync)
12.5 Backoff-until backpressure recovery
12.6 Bidirectional sync conflicts
12.7 Partial failure recovery in multi-peer scenarios
12.8 Replication schedule verification (cron-based replication actually fires)
12.9 Failover between multiple peers
12.10 Cross-instance metadata aggregation (e.g., merging Conda repodata, Helm charts, CRAN PACKAGES across peers)

brandonrc:
## Sub-issue breakdown (Cluster H, in-flight via PR #150)

PR #150 scaffolds `tests/federation/` with two starter tests. All federation tests gate on `PEER1_URL` and skip cleanly in single-instance dev envs (the multi-instance release-gate env is the only one that exercises these paths).

### Covered in PR #150

- 12.1 (scaffold) Peer degradation -> recovery -- `test-peer-degradation-recovery.sh` registers a peer at a closed port and asserts the health probe flips to `degraded` within 30s. The recovery half (re-point URL, wait for `healthy`) is TODO and blocks on PATCH /peers/{id} accepting endpoint_url updates without a full re-registration
- 12.4 (scaffold) Sync filter application -- `test-sync-filter-application.sh` pins the API round-trip for the `filter` field on sync policies. End-to-end "upload .tar.gz + .txt, trigger sync, assert only .tar.gz on peer" is TODO and blocked on artifact-keeper-fzj (sync worker not running in ephemeral env)

### Open sub-issues to file

- 12.1.a Peer recovery -- second half of the degradation/recovery cycle once `PATCH /peers/{id}` accepts endpoint_url
- 12.2 Bandwidth limit enforcement -- needs a controlled-rate transfer test against `max_bandwidth_bps`. Coordinate with whoever owns the sync-worker fix (artifact-keeper-fzj) since this needs the worker actually running
- 12.3 Concurrent transfer queue saturation -- depends on artifact-keeper-fzj
[...]

Issue #77, opened 2026-04-27
Epic 10: Admin operations E2E coverage (v1.1.9)


brandonrc:
Admin and system endpoints in 1.1.x backend that have no E2E coverage. Tracked under coverage policy #66.

## Sub-tasks (1.1.x scope)

10.1 Backup execute (`POST /admin/backups/{id}/execute` -- manually trigger pending backup)
10.2 Backup cancel (`POST /admin/backups/{id}/cancel` -- stop in-progress)
10.3 Backup delete (`DELETE /admin/backups/{id}`)
10.4 `/livez` liveness probe (only `/health` and `/readyz` currently tested)
10.5 Monitoring alerts query (`GET /admin/monitoring/alerts`)
10.6 Alert suppression (`POST /admin/monitoring/alerts/suppress`)
10.7 Manual health check trigger (`POST /admin/monitoring/check`)
10.8 Settings update (`POST /admin/settings`) -- modify and verify config
10.9 List storage backends (`GET /admin/storage-backends`)
10.10 Reindex trigger (`POST /admin/reindex`) and verify search reflects rebuild
10.11 Admin-initiated password reset (`POST /users/{id}/password/reset`)
10.12 User self-service password change (`POST /users/{id}/password`)
10.13 Audit log emission, query, filter, retention, export

brandonrc:
## Sub-issue breakdown (Cluster H, in-flight via PR #150)

PR #150 ships a starter set of admin-operation contract tests under `tests/admin/`. Five subtasks are partially covered with contract pins; the rest remain open as discrete sub-issues:

### Covered in PR #150 (contract-level only)

- 10.1/10.2/10.3 Backup execute/cancel/delete -- `tests/admin/test-backup-lifecycle.sh` (status-code contract; needs an end-to-end "execute -> verify artifact written to storage -> cancel race" follow-up)
- 10.4 `/livez` -- `tests/admin/test-livez.sh` (200 + body-size cap)
- 10.5/10.6/10.7 Monitoring alerts query / suppress / manual check -- `tests/admin/test-monitoring-alerts.sh` (response shape + negative cases)
- 10.9 Storage backends list -- `tests/admin/test-storage-backends-list.sh` (shape + 401/403)
- 10.10 Reindex trigger -- `tests/admin/test-reindex.sh` (status + authz; does not verify search reflects rebuild)

### Open sub-issues to file

[...]

Issue #76, opened 2026-04-27
Epic 11: Auth edge cases E2E coverage (v1.1.9)


brandonrc:
Auth flows in 1.1.x backend that are configured/CRUDed but never run end-to-end with a real client. Tracked under coverage policy #66.

## Sub-tasks (1.1.x scope)

11.1 LDAP full login flow (real or mocked LDAP server, credential validation)
11.2 SAML full login flow (assertion response parsing, signature validation, attribute extraction)
11.3 OIDC state/nonce mismatch attacks
11.4 OIDC redirect URI mismatch detection
11.5 OIDC custom claim mapping (non-standard `username_claim`, `email_claim`, `groups_claim`)
11.6 LDAP group membership sync on federated auth
11.7 Token exchange endpoint (`POST /exchange` used internally by SSO flows)
11.8 API token scope enforcement (read-only token rejects write; admin scope only by admins)
11.9 Download ticket creation, single-use enforcement, expiry
11.10 TOTP backup code consumption and exhaustion
11.11 TOTP disable flow (password + valid TOTP required)
11.12 User deactivation immediately revokes active tokens
11.13 Refresh token age limits and rotation
11.14 Reset password temporary credentials (entropy, one-time use, expiry)
11.15 Password strength validation (unicode, null bytes, length boundaries)

brandonrc:
## Sub-issue breakdown (Cluster H, in-flight via PR #150)

PR #150 adds two new files under `tests/auth/`. The GHSA regression suite (`tests/security/test-ghsa-vvc3-scope-checks.sh`) covers 11.8 end-to-end.

### Covered in PR #150

- 11.8 API token scope enforcement -- `tests/security/test-ghsa-vvc3-scope-checks.sh` plus the existing `tests/auth/test-api-token-scope.sh` and `tests/rbac/test-scope-enforcement.sh`. Read-scope SA tokens rejected with the literal `Token does not have required scope: write` message on the three highest-risk surfaces (permissions, groups, format-handler push paths npm/pypi/oci); admin JWTs continue to pass.
- 11.12 (partial) User deactivation revokes tokens -- new `test-jwt-after-password-change.sh` covers the password-change branch of the same `CREDENTIAL_INVALIDATIONS` map. Existing `test-user-deactivation-revokes-tokens.sh` covers the deactivation branch.
[...]

PR #150, opened 2026-05-15
test: admin + auth + federation E2E coverage with GHSA regression (#77, #76, #78)

brandonrc:
## Summary

Cluster H coverage: admin operations, auth edge cases, and federation depth. Pairs with the GHSA-vvc3-h39c-mrq5 scope-check fix shipped on artifact-keeper main (PR #1219).

### Top priority: GHSA-vvc3-h39c-mrq5 regression test

`tests/security/test-ghsa-vvc3-scope-checks.sh` is the suite that proves the SA-token scope bypass is actually fixed end-to-end:

- A read-scope SA token MUST be rejected (403) on POST /api/v1/permissions, POST /api/v1/groups, npm publish, pypi upload, oci blob upload
- Response body MUST contain `Token does not have required scope: write`
- A read+write SA token MUST succeed on the same endpoints (positive control)
- An admin JWT MUST pass scope checks regardless (JWTs are not scoped; defensive check against over-rotation)

### Admin operations (#77, Epic 10)

`tests/admin/` (new directory, 5 contract tests):

- `test-livez.sh` -- liveness probe contract, unauthenticated 200
- `test-backup-lifecycle.sh` -- execute/cancel/delete (Epic 10.1-10.3)
- `test-monitoring-alerts.sh` -- alerts query, suppression, manual check (Epic 10.5-10.7)
- `test-storage-backends-list.sh` -- list + 401/403 authz pinning (Epic 10.9)
- `test-reindex.sh` -- trigger contract (Epic 10.10)

New `admin-tests` job wired into `.github/workflows/release-gate.yml` (added to `needs:` of `collect-results` and the summary loop / failure-gate condition).

### Auth edge cases (#76, Epic 11)

Two additions to `tests/auth/`:

- `test-malformed-tokens.sh` -- missing header, empty bearer, garbage, three-segment non-JWT, Basic on JWT endpoint, lowercase scheme
- `test-jwt-after-password-change.sh` -- PR #931 CREDENTIAL_INVALIDATIONS regression. Asserts pre-change JWT is rejected (401) after the user rotates their password.

### Federation (#78, Epic 12)

`tests/federation/` (new directory, 2 tests gated on `PEER1_URL`):

- `test-sync-filter-application.sh` -- filter glob round-trip through sync-policy API (Epic 12.4)
- `test-peer-degradation-recovery.sh` -- health-state transition observability (Epic 12.1)

Both document the `artifact-keeper-fzj` constraint (sync worker not running in ephemeral envs) and skip cleanly when the worker is absent.

## Pragmatic scope

The epics are big. This PR ships a focused GHSA regression suite plus 2-5 contract tests per epic. Sub-issue breakdowns posted on each epic so the remaining work is visible.

## Verification

- `bash -n` clean on all 10 new scripts
- `shellcheck -S warning` clean on all 10 new scripts (one `SC2206` suppressed inline with a directive comment; the unquoted expansion of `CURL_TIMEOUT` is intentional and matches the existing pattern in `tests/lib/common.sh`)
- `python yaml.safe_load` parses the updated `release-gate.yml`

### DIFF
--- a/.github/workflows/release-gate.yml
+++ b/.github/workflows/release-gate.yml
@@ -973,7 +973,9 @@ jobs:
   # backend/src/services/auth_service.rs, so any auth-tests step
   # running in parallel that uses the admin JWT starts getting 401s
   # mid-suite once its cached token validation flips to "rejected".
-  # See #137.
+  # See #137. admin-tests is serialized for the same reason and on the
+  # same job ordering (see the comment block above the admin-tests
+  # job definition below).
   # -------------------------------------------------------------------
   platform-tests:
     needs: [deploy, auth-tests]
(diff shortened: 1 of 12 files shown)


brandonrc:
GitGuardian now green after rewriting the branch history into a single commit (commit message records the rewrite, force-push used `--force-with-lease`). All other checks green. Ready for review.

Issue #153, opened 2026-05-15
DTF perf tier: dedicated slot + baseline-relative sustained-load oracle (make stress gating)


brandonrc:
## Context

`tests/stress/test-sustained-load.sh` currently asserts `error_pct <= 55%` (down from a previous 30% that deterministically failed on v1.1.9-rc.1). The release-gate job runs with `continue-on-error: true`, so the assertion is informational, but the 55% number is a fixed ceiling chosen from a tiny sample (worst observed run 53% plus margin).

A real regression that produces 40% error rate would slip past, because the threshold is set to absorb the worst observed flake.

## Followup from PR #148

PR #148 (Fresh-Eyes review, Finding 1) called this out. The original review asked for either tightening the threshold or making it telemetry-only. We kept the assertion plus `continue-on-error: true` for now because the suite is genuinely flake-prone on shared ARC runners and a hard fail would block release-gate without giving signal.

## Proposal

Once 5-10 successful baseline runs have accumulated on the current Helm CPU layout (post-#140 rebalance, post-#132 admin-exempt credential), record their error-rate distribution and switch the assertion to:

```
error_pct <= median(baseline_error_pct) + 15
```

If a `tests/stress/baseline-sustained-error-pct.txt` file is absent, the script should warn loudly and pass (instead of failing with "no baseline" as the original review suggested) so the release-gate stays unblocked while the baseline file is being populated.

## Acceptance

- Baseline file checked in with at least 5 successful run values
- Script computes median + delta at runtime, not a hard-coded number
- `SUSTAINED_ERROR_PCT_THRESHOLD` env override still works for dedicated-runner environments
- TODO comment in `test-sustained-load.sh` removed when this lands

brandonrc:
## Status (2026-05-16)

This issue is a followup on Finding 1 from the Fresh-Eyes review of PR #148 (stress retry budgets / Helm CPU rebalance / sustained-load tuning). The proposed switch to a baseline-relative threshold needs:

1. PR #148 to land so the new Helm CPU layout is the canonical baseline.
2. 5-10 sustained-load runs against that layout so the median/distribution is meaningful.

Both prerequisites are external to this issue. Recommend leaving this open as a v1.2.0 followup that re-opens once #148 has been merged and 5-10 release-gate runs have accumulated.

brandonrc:
**Keeping this open and reframing it as the perf/load tier work item.** Retitled,
[...]

PR #162, opened 2026-05-16
ci(stress): wire collect-pg-stats into stress-tests job (#154)

### LINKED ISSUE #154: stress: capture pg_stat_activity / acquire-time evidence on sustained-load runs
## Context

PR #140 rebalanced postgres CPU limits in `helm/values-test.yaml` and `helm/values-test-full.yaml` based on the narrative that the previous 5:1 backend:postgres ratio was saturating the sqlx pool and surfacing as the 33-54% sustained-load error rate.

The narrative is plausible but is not currently backed by direct measurement in the test run. Fresh-Eyes review on PR #148 (Finding 2) flagged this as "narrative not measurement" and recommended either softening the claim or capturing the evidence.

PR #148 softened the claim (the commit body now says "should reduce" rather than "fixes") and added a `tests/stress/collect-pg-stats.sh` capture script + workflow artifact upload, but does not yet wire the capture step into the stress-tests job. This issue tracks doing that wire-up.

## Proposal

1. Add a `Collect postgres stats` step to the `stress-tests` job in `.github/workflows/release-gate.yml`, running `tests/stress/collect-pg-stats.sh` against the namespace before teardown.
2. The script should snapshot:
   - `SELECT * FROM pg_stat_activity WHERE state != 'idle'`
   - `SELECT * FROM pg_stat_statements ORDER BY total_exec_time DESC LIMIT 20` (if extension is available)
   - `kubectl top pod -n <ns>` for backend + postgres pods
3. Upload the snapshot as a workflow artifact (`stress-pg-stats`).
4. Investigate sqlx `pool_acquire_time` histogram exposure on the backend side (likely an artifact-keeper backend change, file there if needed).

## Acceptance

- Stress-tests run produces a `stress-pg-stats` artifact alongside `stress-request-logs`
- Artifact contains pg_stat_activity rows + kubectl top output for at least one run
- If sqlx instrumentation is exposed, also include the histogram dump

brandonrc:
## Summary

Wires the `tests/stress/collect-pg-stats.sh` capture script (landed in #148) into the `stress-tests` job of `.github/workflows/release-gate.yml`, so every release-gate run produces direct measurement of postgres state alongside `stress-request-logs`.

Background: PR #140 rebalanced postgres CPU limits on the narrative that the previous 5:1 backend:postgres ratio was saturating the sqlx pool. Fresh-Eyes review on PR #148 (Finding 2) flagged this as "narrative not measurement". PR #148 added the capture script but did not wire it into CI. This PR finishes that wire-up.

## Changes

- Added two steps to the `stress-tests` job, after the test run and before teardown:
  - `Collect postgres stats` (`if: always()`) invokes `bash tests/stress/collect-pg-stats.sh`.
  - `Upload pg-stats snapshot` (`if: always()`) uploads `/tmp/pg-stats/` as `stress-pg-stats-${{ github.run_attempt }}` using the same SHA-pinned `actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a` (v7.0.1) the rest of the workflow uses.
- Added `PG_STATS_DIR=/tmp/pg-stats` and `NAMESPACE=${{ needs.deploy.outputs.namespace }}` to the job's `env:` block. These are the two env vars the capture script reads (see header comment in `tests/stress/collect-pg-stats.sh`). The script does not consume a separate run-id variable; the run id is folded into the artifact name via `${{ github.run_attempt }}`.
- No changes to the capture script itself (respects #148), no changes to other jobs in the workflow.

Both new steps use `if: always()` so the snapshot ships even when the stress workload fails, which is exactly when the `pg_stat_activity` / connection-count / `kubectl top` data is most useful for triage.

## Acceptance (per #154)

- [x] Stress-tests run produces a `stress-pg-stats` artifact alongside `stress-request-logs`.
- [x] Artifact will contain `pg_stat_activity` rows + `kubectl top` output (the capture script writes these; the wire-up ensures the directory ships).
- [ ] sqlx `pool_acquire_time` histogram dump - tracked separately as a backend (artifact-keeper) change per the issue body; out of scope for this test-repo PR.

Diff: 1 file changed, 29 insertions(+), 0 deletions(-).

Closes #154

## Test plan

- [ ] Confirm the new steps render in the next release-gate workflow run.
- [ ] Inspect the `stress-pg-stats-<run_attempt>` artifact for `_meta.txt`, `kubectl-top-pods.txt`, `pg_stat_activity.txt`, `pg_connection_count.txt`.
- [ ] Verify the upload step still fires when the `Run stress tests` step exits non-zero (force a failure locally or wait for a real failing run).

### DIFF
--- a/.github/workflows/release-gate.yml
+++ b/.github/workflows/release-gate.yml
@@ -1509,6 +1509,15 @@ jobs:
       # aggregate error counts alone (artifact-keeper-test#138 /
       # artifact-keeper#1088).
       STRESS_LOG_DIR: /tmp/stress-logs
+      # Postgres + pod-resource snapshot directory. The Collect postgres
+      # stats step below runs tests/stress/collect-pg-stats.sh after the
+      # stress run and ships this directory as the stress-pg-stats
+      # artifact. Captures the direct measurement that PR #148's
+      # Fresh-Eyes review (artifact-keeper-test#154) asked for: pg
+      # connection saturation + kubectl top, so the postgres-CPU
+      # narrative behind PR #140 is backed by evidence on every run.
+      PG_STATS_DIR: /tmp/pg-stats
+      NAMESPACE: ${{ needs.deploy.outputs.namespace }}
       # RELEASE_GATE=1 turns common.sh's skip_suite() into a hard fail.
       # A silently-skipped test in release-gate context is exactly the
       # silent-success class (#870/#871/#888) the gate exists to catch.
@@ -1564,6 +1573,26 @@ jobs:
               | sort | uniq -c | sort -rn | head -n 10 \
               | sed 's/^/    /'
           done
+
+      - name: Collect postgres stats
+        # Snapshot pg_stat_activity, pg_stat_statements, connection-count
+        # vs max_connections, and kubectl top for the backend + postgres
+        # pods. Runs after the stress workload and before teardown so the
+        # capture reflects steady-state load. if: always() ensures the
+        # snapshot still ships on a failed run, which is exactly when
+        # the data is most useful (artifact-keeper-test#154; rationale
+        # in PR #148 Fresh-Eyes review, Finding 2).
+        if: always()
+        run: |
+          bash tests/stress/collect-pg-stats.sh
+
+      - name: Upload pg-stats snapshot
+        if: always()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: stress-pg-stats-${{ github.run_attempt }}
+          path: /tmp/pg-stats/
+          if-no-files-found: ignore
 
       - name: Upload test results
         if: always()


brandonrc:
## Backend / Observability Review

**1. Reach a running pg pod before teardown:** ✅ Teardown lives in a separate `teardown` job (`release-gate.yml:1903`), so it cannot start until `stress-tests` finishes. The collect step (`release-gate.yml:1577`) runs in-job, before the stress-tests job exits, so the namespace is guaranteed alive at capture time. `if: always()` (`release-gate.yml:1585`, `1590`) covers the crashed-backend case — `pg_stat_activity` + `kubectl top` still ship.

**2. pg_stat_statements coverage:** ⚠ The extension is opt-in via `shared_preload_libraries`, and `grep -rn shared_preload\|pg_stat_statements helm/values-test*.yaml` returns zero hits — the test deployment does **not** preload it. The script's fallback at `collect-pg-stats.sh:115` catches this and writes `(pg_stat_statements extension is not enabled in this build)` rather than an empty file. Acceptable, but consider a follow-up to enable the extension in `values-test.yaml` so probe 5 is more than a placeholder.

**3. Sampling cadence:** ⚠ Single end-of-run snapshot, matching #154's stated design. Will miss the peak if it happens mid-run and recovers. Acceptable for #154's "evidence on every run" bar; flag as a known limitation if the data is later used to justify another CPU rebalance.

**4. Retention:** ✅ GitHub default 90 days is sufficient for the 5-10-run baseline #153 calls for.

**5. Scope:** ✅ Out-of-scope `sqlx pool_acquire_time` histogram is explicitly deferred in the PR body (artifact-keeper change), and the diff confirms — zero backend changes here.

**6. Diff hygiene:** ✅ Surgical: 29 additions, 0 deletions, 1 file. Two new steps + 2 env vars (`PG_STATS_DIR`, `NAMESPACE` at `release-gate.yml:1517,1520`) — exactly the env the script reads. No neighbor refactors. SHA-pinned `upload-artifact@043fb46d…` matches the rest of the workflow.

**Verdict:** Approve. One non-blocking follow-up: enable `pg_stat_statements` in `helm/values-test.yaml` so probe 5 produces real data.

— Backend Architect agent

brandonrc:
## Independent review

Clean, minimal wire-up. Reads correctly against the script contract in `tests/stress/collect-pg-stats.sh` and the deploy job's outputs. Recommending ship.

- PASS - Step placement: `Collect postgres stats` (release-gate.yml:1577) runs after `Run stress tests` / `Summarize per-request status codes` and before any teardown, inside the same `stress-tests` job. Not a separate job, so it shares the runner's kubeconfig and the deploy job's outputs.
- PASS - `if: always()` is set on both the collect step (line 1585) and the upload step (line 1590). A failed stress run still ships the snapshot, which matches the "data is most useful on failure" rationale.
- PASS - `NAMESPACE` (line 1520) is sourced from `needs.deploy.outputs.namespace`, which is declared at release-gate.yml:377. `PG_STATS_DIR=/tmp/pg-stats` (line 1519) matches the script's default and is the exact path the upload step references at line 1594. Both env vars match the contract documented in the script header (collect-pg-stats.sh:20-22).
- PASS - SHA pin matches the repo convention in `docs/CONTRIBUTING.md` lines 3-7: `actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1` is identical to the 20 other upload-artifact references in this workflow. No floating tags introduced.
- PASS - Upload path glob: `path: /tmp/pg-stats/` catches everything the script writes (`_meta.txt`, `kubectl-top-pods.txt`, `kubectl-get-pods.txt`, `pg_stat_activity.txt`, `pg_stat_statements.txt`, `pg_connection_count.txt`, `backend-pool-lines.txt`, plus the `SKIPPED.txt` / `POSTGRES-NOT-FOUND.txt` sentinel files on the bail-out paths). `if-no-files-found: ignore` is the right call given the script exits 0 on missing kubectl / namespace.
- PASS - No collision with `stress-request-logs`: different artifact name (`stress-pg-stats-<run_attempt>` vs `stress-request-logs`) and different path (`/tmp/pg-stats/` vs `/tmp/stress-logs/`). The new artifact's name also includes `${{ github.run_attempt }}`, matching nothing else in the workflow.
- PASS - YAML still parses: `python3 -c \"import yaml; yaml.safe_load(open('.github/workflows/release-gate.yml'))\"` returns clean.
- PASS - No scope creep. `git diff main --stat` shows a single file touched, 29 insertions, 0 deletions. No refactors, no unrelated jobs modified, no changes to the capture script itself.

Nit (non-blocking, do not hold the PR): the `stress-request-logs` artifact at line 1615 does not include `${{ github.run_attempt }}` in its name, while the new `stress-pg-stats` one does. Minor inconsistency, but the new name is the better pattern (re-runs do not collide). Worth normalizing in a follow-up, not here.

- Code Reviewer agent

PR #164, opened 2026-05-16
test(search): tree browse + sort + pagination/autocomplete edges + checksum search (#73)

brandonrc:
## Summary

Closes a chunk of issue #73 (Epic 8: search depth) with six new E2E scripts under `tests/search/`. All scripts are auto-discovered by `./scripts/run-suite.sh --suite search` and live alongside the existing search tests as siblings.

### Subtasks delivered

- [x] **8.6** `sort_by` / `sort_order` parameters on `/api/v1/search/advanced` (`test-search-sort.sh`). Load-bearing assertion: `sort_order=asc` and `sort_order=desc` on `sort_by=size` produce different first hits; supported and unsupported sort fields/orders must not 5xx.
- [x] **8.10** Tree browse via the OpenAPI-canonical query-string surface `/api/v1/tree?repository_key=&path=` (`test-search-tree-browse-query.sh`). Complements the existing path-routing test by exercising the documented form, plus 4xx-on-missing-repo and `include_metadata=true` smoke.
- [x] **8.11** Pagination overflow edges (`test-search-pagination-overflow.sh`). Adds the cases the existing `test-search-pagination.sh` does not cover: `per_page=-1` (signed-to-unsigned underflow), `page=2147483647` (offset arithmetic overflow), `page=0`+`per_page=0` (combined underflow), `per_page=100000` (clamp / DoS surface).
- [x] **8.12** Autocomplete `/api/v1/search/suggest` edges (`test-search-autocomplete-edges.sh`): empty prefix, 4 KiB prefix, Lucene/regex metacharacter prefix, `limit=0`, `limit=-1`, plus a sanity-hit assertion so a no-op handler can not pass.
- [x] **8.13** Checksum search across SHA1 + MD5 (`test-search-checksum-multi.sh`). The existing `test-search-checksum.sh` only exercises SHA256. Load-bearing assertion: SHA1 and MD5 lookups return the **same artifact id** as the SHA256 baseline; this catches a silent-default regression where the `algorithm` parameter is parsed but ignored. Also asserts 422 for unsupported algorithms and case-insensitive lookup.
- [x] **8.15** Recent endpoint custom `limit` parameter (`test-search-recent-limit.sh`): `limit=1` and `limit=3` clamp respected; `limit=0` must produce an empty array (not the default page); `limit=-1` no 5xx; `limit=10000` clamped or rejected.

### Deferred to follow-up PRs

- [ ] **8.7** Tag/label-based search (`artifact_labels` integration) - needs label-fixture infra that does not exist yet.
- [ ] **8.8** Faceted drill-down (filter -> facet-refine; verify counts match filtered results) - multi-step, larger payload than this PR.
- [ ] **8.9** Incremental indexing pipeline SLA (upload -> search returns within SLA) - timing-sensitive; flake risk on parallel runners.
- [ ] **8.1 / 8.2 / 8.3 / 8.4 / 8.5** Already covered by existing `test-search-advanced-filters.sh` (size, date, repository_key) and untouched here.
- [ ] **8.14** Trending `days` and `limit` parameters - same shape as 8.15, can pile on as a follow-up alongside 8.7.

### Design notes

- Every script gates on a preflight probe and `skip_suite`s cleanly (not silently passes) when the endpoint is unmounted (404/501) or the search backend is unavailable (503/504/000).
- Every script uses `RUN_ID` in all resource names and registers an EXIT-trap cleanup via `add_exit_handler`.
- The overflow tests use ONLY non-5xx as the load-bearing contract; exact 4xx-vs-2xx behaviour varies across backend revisions and is intentionally not asserted.
- `tests/lib/common.sh` is not modified.
- The release-gate `search-tests` job invokes `./scripts/run-suite.sh --suite search`; new files in `tests/search/` are auto-discovered, so `.github/workflows/release-gate.yml` is not touched.

## Test plan

- [ ] CI `search-tests` job runs all six new scripts against the test namespace.
- [ ] Each new script prints PASS / SKIP / FAIL per case and writes its JUnit XML to `$JUNIT_OUTPUT_DIR/`.
- [ ] `bash -n` syntax-clean on all six (verified locally).
- [ ] No credentials embedded; GitGuardian clean.

Refs #73

### DIFF
--- a/tests/search/test-search-autocomplete-edges.sh
+++ b/tests/search/test-search-autocomplete-edges.sh
@@ -1,0 +1,201 @@
+#!/usr/bin/env bash
+# test-search-autocomplete-edges.sh - Autocomplete (/search/suggest) edges
+#
+# Epic 8 sub-task 8.12 (artifact-keeper-test#73). The suggest endpoint
+# (OpenAPI: GET /api/v1/search/suggest?prefix=...&limit=...) has only a
+# smoke probe today (test-search-basic.sh checks 2xx). Edge cases that
+# have caused real regressions in this handler:
+#
+#   1. Empty prefix (prefix="") -> the OpenAPI marks prefix as required.
+#      Must reject with 4xx; a 5xx would imply the handler tried to
+#      walk a zero-length string.
+#
+#   2. Very long prefix (4 KiB) -> tests the prefix-length cap. Must
+#      not 5xx; reasonable behaviours are 4xx (rejected) or 2xx with
+#      no suggestions.
+#
+#   3. Prefix containing special characters (Lucene/regex metacharacters,
+#      Unicode, percent-encoded bytes) -> must not 5xx, must not echo
+#      user input back as part of a server-side error.
+#
+#   4. limit=0 -> must produce an empty suggestions array (not a panic,
+#      not the default-limited list silently).
+#
+#   5. limit=-1 -> must 4xx or be coerced; must not 5xx.
+#
+#   6. Happy path: prefix matching our sentinel returns >= 1 suggestion.
+#      This is the "we actually exercise the indexer" sanity check.
+#
+# Skips cleanly if /api/v1/search/suggest is not mounted (404/501).
+#
+# Requires: curl, jq
+
+source "$(dirname "$0")/../lib/common.sh"
+
+begin_suite "search-autocomplete-edges"
+auth_admin
+setup_workdir
+
+# Preflight: the OpenAPI says prefix is required, so a probe without it
+# may legitimately 4xx; treat that as "endpoint exists" too.
+preflight_status=$(curl -s -o /dev/null -w '%{http_code}' --max-time 5 \
+  -H "$(auth_header)" \
+  "${BASE_URL}/api/v1/search/suggest?prefix=preflight" 2>/dev/null || echo "000")
+case "$preflight_status" in
+  404|501)
+    skip_suite "suggest endpoint not available (HTTP ${preflight_status})"
+    ;;
+  503|504|000)
+    skip_suite "search backend unavailable (HTTP ${preflight_status})"
+    ;;
+esac
+
+REPO_KEY="autocp-${RUN_ID}"
+UNIQUE_TERM="acprfx${RUN_ID//[^a-z0-9]/}"
+
+add_exit_handler "api_delete \"/api/v1/repositories/${REPO_KEY}\" >/dev/null 2>&1 || true"
+
+begin_test "Create repo and upload sentinel for suggestions"
+if create_local_repo "$REPO_KEY" "generic" >/dev/null 2>&1; then
+  echo "autocomplete-sentinel-${UNIQUE_TERM}" > "${WORK_DIR}/${UNIQUE_TERM}.txt"
+  api_upload "/api/v1/repositories/${REPO_KEY}/artifacts/${UNIQUE_TERM}/${UNIQUE_TERM}.txt" \
+    "${WORK_DIR}/${UNIQUE_TERM}.txt" >/dev/null 2>&1 || true
+  pass
+else
+  fail "could not create repo"
+  end_suite
+fi
+
+sleep 3   # allow autocomplete index to settle
+
+# _suggest_status <query-string>
+# Returns HTTP status for a suggest request.
+_suggest_status() {
+  local qs="$1"
+  curl -s -o /dev/null -w '%{http_code}' $CURL_TIMEOUT \
+    -H "$(auth_header)" \
+    "${BASE_URL}/api/v1/search/suggest?${qs}" 2>/dev/null || echo "000"
+}
+
+# _suggest_count <query-string>
+# Returns the number of suggestions in a 2xx response, or "" on error.
+_suggest_count() {
+  local qs="$1"
+  local body
+  body=$(curl -sf $CURL_TIMEOUT -H "$(auth_header)" \
+    "${BASE_URL}/api/v1/search/suggest?${qs}" 2>/dev/null || echo "")
+  if [ -z "$body" ]; then
+    echo ""
+    return
+  fi
+  echo "$body" | jq -r '
+    if (.suggestions? | type) == "array" then (.suggestions|length)
+    elif type == "array" then length
+    elif (.items? | type) == "array" then (.items|length)
+    else 0 end' 2>/dev/null || echo ""
+}
+
+# -------------------------------------------------------------------------
+# 1. Empty prefix (prefix="") must not 5xx. 4xx is the documented response.
+# -------------------------------------------------------------------------
+
+begin_test "prefix= (empty) does not 5xx"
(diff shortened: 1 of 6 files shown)


brandonrc:
## Code Review — PR #164 (Epic 8 search subtasks)

**Overall:** strong submission. Scope is clean (6 files under `tests/search/`, no creep), no secrets, `RUN_ID` used in every resource name, EXIT-trap cleanup wired via `add_exit_handler`, and helpers (`api_get`, `auth_admin`, `begin_suite`, etc.) used correctly. No subtask is double-covered — I diffed against `test-search-pagination.sh`, `test-search-tree-browse.sh`, `test-search-advanced-filters.sh`, `test-search-checksum.sh`, `test-search-basic.sh` and the new probes are disjoint.

### Blockers (none)

### Suggestions

- ⚠ `test-search-autocomplete-edges.sh:102-107` — empty-prefix test only checks non-5xx. Review focus #2 specifically asks: it must NOT silently return everything. Add a 2xx-branch assertion that the suggestions array is either empty or short (e.g. `<= limit`), otherwise a backend that ignores `prefix=""` and dumps the index passes this test. Mirror the shape used at `:148-165` for `limit=0`.
- ⚠ `test-search-pagination-overflow.sh:42-49` — preflight only skips on 503/504/000; a 404 on `/api/v1/search` will silently fall through to the per-endpoint discovery loop at `:71-77`, which is fine, but the suite then runs 4 probes that all `skip_suite` via `:84`. Consider adding `404|501) skip_suite ...` to the preflight `case` for symmetry with the other 5 new files.
- ⚠ `test-search-sort.sh:79`, `test-search-recent-limit.sh:67`, `test-search-autocomplete-edges.sh:69` — fixed `sleep 3-4` for async indexing is fragile on slow runners. The pattern at `test-search-autocomplete-edges.sh:191-199` (probe-then-extra-sleep) is better; consider lifting it into a helper.
- ⚠ `test-search-checksum-multi.sh:188` — accepts `400|404` alongside `422` as "acceptable rejection". OpenAPI says 422 — if the contract is load-bearing, tighten to `422` only and `fail` on `400/404`. If it's intentionally permissive, the comment at `:187` should say so.

### Nits

- 💭 `test-search-sort.sh:104-105` — the `_first_name` helper is robust, but if the backend ever returns identical names with different versions, asc/desc could legitimately match. Worth a comment that the three sentinels have distinct basenames by construction (small/med/large).
- 💭 `test-search-pagination-overflow.sh:174` — clamp threshold of 1000 is generous; the inline comment is good. Same threshold reused at `test-search-recent-limit.sh:175` — consider a shared constant.
- 💭 `test-search-checksum-multi.sh:200-214` — case-insensitive test is a nice catch beyond the subtask scope.

### What's good

- Load-bearing assertions are explicit where it counts: `:132-136` (INT32_MAX returns 0 hits, not silent coercion), `:148-151` (SHA1 id matches SHA256 id — catches silent-default regression), `:114-118` (sort asc/desc head must differ), `:131-147` of tree-browse-query (path param actually applied).
- Skip-not-pass discipline is clean throughout; every `skip` has a reason string.
- Excellent file-header comments explaining the regression each probe targets.

— Code Reviewer agent

brandonrc:
## Backend Architect review (correctness lens)

Verified each test against `artifact-keeper-api/openapi.yaml` (authoritative).

**Endpoint/param fidelity vs OpenAPI**
- OK `test-search-sort.sh:104-105` uses `/api/v1/search/advanced?query=&sort_by=&sort_order=&per_page=` matching openapi.yaml:7810-7919. The tolerant `_first_name` extractor (lines 85-97) handles `.items` (the documented shape per `AdvancedSearchResponse` at 10173) plus legacy wrappers.
- OK `test-search-tree-browse-query.sh:112,132` uses `/api/v1/tree?repository_key=&path=&include_metadata=` matching openapi.yaml:9272-9308. `TreeResponse.nodes[]` shape (17772-17780) honored.
- OK `test-search-pagination-overflow.sh:71-77` correctly notes `/api/v1/search` is not in OpenAPI and falls back to `/api/v1/search/advanced`. Sends both `q=` and `query=` (line 89) which is defensive but harmless.
- OK `test-search-autocomplete-edges.sh` uses `/api/v1/search/suggest?prefix=&limit=` matching openapi.yaml:8012-8037.
- OK `test-search-checksum-multi.sh` uses `/api/v1/search/checksum?checksum=&algorithm=` matching openapi.yaml:7927-7953; cross-checks `.artifacts[0].id` (ChecksumArtifact.id is required, 11308-11320).
- OK `test-search-recent-limit.sh` uses `/api/v1/search/recent?limit=` matching openapi.yaml:7989-8011, returning bare `SearchResultItem[]` (16784).

**Eventual-consistency / Meilisearch lag**
- OK Every test that depends on an indexed write has a `sleep 2-4` after upload (sort:79, tree:105, pagination-overflow:67, autocomplete:69, checksum:110, recent:67) AND the autocomplete sanity check has a 4s retry path (test-search-autocomplete-edges.sh:191-198). No immediate-consistency assumptions found.

**Contract pinning correctness**
- OK Pagination tests assert ONLY non-5xx for `page=0`, `per_page=-1`, `page=INT32_MAX`. OpenAPI says `minimum: 0` for both (7896, 7905) so negative values are out-of-spec; accepting either 4xx or 2xx-ignore is the right shape.
- OK Autocomplete empty `prefix=`: OpenAPI marks `prefix` as required (8020) so 4xx is correct; the test only forbids 5xx (line 103-107), letting 2xx-empty pass too. Looser than the spec demands but not wrong.
- WARN `test-search-checksum-multi.sh:200-214` asserts case-insensitive uppercase-SHA1 lookup as a hard `fail`. That behavior is NOT in OpenAPI (no normalization clause on `checksum` param at 7933). If the backend treats the param as case-sensitive, this test flips red even though the backend is spec-compliant. Recommend downgrading to `skip` or documenting the contract elsewhere.
- WARN `test-search-checksum-multi.sh:188` accepts `400|404` alongside `422` for unsupported algorithm; OpenAPI explicitly documents 422 (7952). Loosening this hides a spec violation. Suggest hard-pinning 422.

**Tree-browse scope honesty**
- OK `test-search-tree-browse-query.sh` covers the load-bearing path: repo-scoped vs path-scoped responses must differ (lines 111-147). Catches the "path parameter ignored" regression. `include_metadata` is best-effort skip (203-205) — appropriate for the 709-line handler since deeper coverage belongs in unit tests.

**v1.2.0 coupling**
- OK No feature dependency beyond what openapi.yaml documents today. All tolerant extractors and skip preflights mean older backends degrade to skip rather than fail.

**Minor**
- INFO `test-search-pagination-overflow.sh:158-183` accepts `per_page=100000` returning up to 1000 rows as "clamped". OpenAPI has no documented max; if the backend later picks 5000 as its cap this test fails. Consider raising the threshold or asserting only on a known-doc maximum.

Net: mergeable. Two checksum assertions (case-insensitivity, 422 broadening) are the only items worth tightening before this becomes a contract reference.

-- Backend Architect agent

Issue #75, opened 2026-04-27
Epic 7: Webhooks, events, and gRPC E2E coverage (v1.1.9)


brandonrc:
Webhook delivery engine, event taxonomy, and gRPC services exist in 1.1.x backend but lack E2E coverage. Tracked under coverage policy #66.

## Sub-tasks (1.1.x scope)

7.1 Webhook delivery retry engine (exponential backoff 30s -> 2m -> 15m -> 1h -> 4h)
7.2 Dead-letter queue handling for exhausted attempts
7.3 Max-attempts exhaustion scenario
7.4 Async retry job triggering (`process_webhook_retries`)
7.5 HMAC signature generation and `X-Webhook-Signature` header injection
7.6 Webhook secret rotation/revocation
7.7 Custom header injection
7.8 SSRF prevention on webhook URLs (private IPs, metadata servers, loopback)
7.9 URL re-validation at delivery time (DNS rebinding prevention)
7.10 Repository-scoped webhook filtering
7.11 Multi-event filtering (currently only artifact.uploaded tested)
7.12 Webhook test/dry-run endpoint (`POST /{id}/test`)
7.13 Webhook enable/disable toggle (`POST /{id}/enable`, `/{id}/disable`)
7.14 Event delivery list filtering by status
7.15 SSE event stream consumption from a client (with backpressure / lagged-event handling)
7.16 Cross-handler event propagation (artifact.created -> webhook delivery -> SSE broadcast)
7.17 gRPC SbomService (port 9090): GenerateSbom, GetSbom, ListSbomsForArtifact, ConvertSbom, DeleteSbom, RegenerateSbom, CheckLicenseCompliance
7.18 gRPC CveHistoryService: GetCveHistory, UpdateCveStatus, GetCveTrends, TriggerRetroactiveScan
7.19 gRPC SecurityPolicyService: license policy CRUD

brandonrc:
## Sub-issue breakdown (proposed)

PR #149 ships three proof-of-concept tests against this epic (custom headers, multi-event filter, SSE stream). Each remaining sub-task below should become its own sub-issue so the work can be parallelised and the epic stays an aggregator rather than a single-PR mega-task.

### Webhook delivery engine

[...]

PR #165, opened 2026-05-16
test(webhooks): HMAC signing + SSRF + dry-run + enable/disable toggle (#75)

brandonrc:
## Summary

Adds five E2E test scripts under `tests/webhooks/` covering well-defined
sub-tasks from issue #75 that were not yet exercised end-to-end. Each
test follows the existing pattern (sources `lib/common.sh`, uses
`RUN_ID`, EXIT-trap cleanup, skips cleanly on missing endpoints, never
silently passes).

## Delivered (issue #75)

- [x] **7.5 HMAC signing (negative path)** — `test-webhook-hmac-tamper.sh`
  Existing `test-webhook-hmac-signature.sh` proves happy-path:
  header == HMAC(secret, t.body). This suite proves the header
  DIVERGES under (a) wrong secret, (b) tampered body, (c) mutated
  timestamp. Without these the happy-path test would pass against a
  backend that emits a static or insufficiently-bound signature.

- [x] **7.8 SSRF prevention on webhook URLs** — `test-webhook-ssrf-prevention.sh`
  Asserts POST `/api/v1/webhooks` with each of these is rejected 4xx:
  cloud metadata (169.254.169.254, metadata.google.internal),
  link-local 169.254.x, IPv4 loopback (127.0.0.1 + alias 127.1.2.3),
  IPv6 loopback ([::1]), `localhost` hostname, RFC1918 (10/8, 172.16/12,
  192.168/16), 0.0.0.0, and non-http schemes (file://, gopher://).
  Load-bearing assertion: a 2xx accept here would let a tenant
  exfiltrate IAM creds or scan internal services.

- [x] **7.12 Webhook dry-run** — `test-webhook-dry-run.sh`
  POST `/{id}/test` returns 2xx, receiver actually got a POST,
  delivered body has `.event`, carries `X-ArtifactKeeper-Delivery`,
  and `/test` on an unknown id returns 404. Catches the silent-
  success regression where `/test` 200s without firing.

- [x] **7.13 Enable/disable toggle** — `test-webhook-enable-disable-toggle.sh`
  POST `/{id}/enable` and `/{id}/disable` flip the persisted
  `.enabled` flag, are idempotent (second call still 2xx), and
  `/enable` on a fabricated id returns 404. Existing CRUD test
  only PUTs `{enabled:false}` and never asserts state.

- [x] **7.14 Delivery status filter** — `test-event-delivery-status-filter.sh`
  GET `/{id}/deliveries?status=<x>`. Drives a delivery, reads
  the observed status vocabulary, asserts filter=<observed>
  returns only matching rows AND filter=<bogus> returns zero rows.
  The bogus-filter assertion is the one that catches "param is
  silently ignored" — without it, an ignored filter passes the
  positive test trivially.

## Explicitly deferred (out of scope for this PR)

- [ ] 7.1 retry engine backoff timing (30s -> 2m -> 15m -> 1h -> 4h)
- [ ] 7.2 dead-letter queue handling for exhausted attempts
- [ ] 7.3 max-attempts exhaustion scenario
- [ ] 7.4 async retry job (`process_webhook_retries`)
  Reason: all four are timing-sensitive and need a larger fixture
  (sustained backend uptime across multi-hour backoff windows).
  Better landed in a dedicated retry-engine PR with its own
  `TEST_TIMEOUT` overrides.

- [ ] 7.16 cross-handler event propagation
  (artifact.created -> webhook delivery -> SSE broadcast)
  Reason: spans the producer + SSE subsystems, deserves its own
  multi-subsystem orchestration suite.

- [ ] 7.17 gRPC SbomService
- [ ] 7.18 gRPC CveHistoryService
- [ ] 7.19 gRPC SecurityPolicyService
  Reason: different transport layer (port 9090, gRPC); the test
  runner currently only wires HTTP fixtures, gRPC needs separate
  fixture infra.

## Test plan

- [ ] CI `webhook-tests` job (release-gate workflow,
      `./scripts/run-suite.sh --suite webhooks`) goes green against
      a deployment of the in-flight backend.
- [ ] Locally:
      `BASE_URL=... ADMIN_USER=... ADMIN_PASS=... RUN_ID=local bash tests/webhooks/test-webhook-hmac-tamper.sh`
      and the four siblings; each emits PASS rows and writes
      `$JUNIT_OUTPUT_DIR/<suite>.xml`.
- [ ] Files chmod +x and `bash -n` clean.

### DIFF
--- a/tests/webhooks/test-event-delivery-status-filter.sh
+++ b/tests/webhooks/test-event-delivery-status-filter.sh
@@ -1,0 +1,276 @@
+#!/usr/bin/env bash
+# test-event-delivery-status-filter.sh
+#
+# Issue #75 sub-task 7.14: GET /api/v1/webhooks/{id}/deliveries supports
+# a `status` query parameter (openapi.yaml: parameter `status`, type
+# nullable string). No existing test exercises the filter; the delivery
+# and dead-letter suites read the list without a filter and inspect every
+# row.
+#
+# The .status field is what the operator UI uses to surface "all
+# failures" / "all retries" / "all successes". A regression where the
+# query parameter is silently ignored returns the unfiltered list, the
+# UI shows the wrong rows, and on-call triages the wrong delivery.
+#
+# Strategy:
+#   1. Drive a SUCCESS delivery via /test against a 200-always receiver.
+#   2. Read the unfiltered deliveries list to learn the actual status
+#      vocabulary the backend uses (it varies across builds:
+#      "success" | "succeeded" | "delivered" are all in the wild).
+#   3. Filter by that observed status. The returned set MUST be
+#      non-empty AND every row's .status MUST equal the queried value.
+#   4. Filter by a status vocabulary that does not match any row
+#      ("xfail-${RUN_ID}") -- the returned set MUST be empty. This is
+#      the load-bearing assertion: if the backend ignores the param,
+#      this query returns the full list and the test fails loudly.
+#
+# Requires: curl, jq, python3
+
+source "$(dirname "$0")/../lib/common.sh"
+
+begin_suite "event-delivery-status-filter"
+
+WEBHOOK_RECEIVER_PORT="${WEBHOOK_RECEIVER_PORT:-18783}"
+WEBHOOK_RECEIVER_URL="${WEBHOOK_RECEIVER_URL:-http://127.0.0.1:${WEBHOOK_RECEIVER_PORT}/hook}"
+WEBHOOK_RECEIVER_LOG="${WEBHOOK_RECEIVER_LOG:-/tmp/mock-webhook-receiver-statusfilter-${RUN_ID}.log}"
+RECEIVER_PID=""
+WEBHOOK_ID=""
+
+cleanup_and_finalize() {
+  local code=$?
+  if [ -n "${WEBHOOK_ID}" ] && [ "${WEBHOOK_ID}" != "null" ]; then
+    api_delete "/api/v1/webhooks/${WEBHOOK_ID}" >/dev/null 2>&1 || true
+  fi
+  if [ -n "${RECEIVER_PID}" ] && kill -0 "${RECEIVER_PID}" 2>/dev/null; then
+    kill "${RECEIVER_PID}" 2>/dev/null || true
+    wait "${RECEIVER_PID}" 2>/dev/null || true
+  fi
+  rm -f "${WEBHOOK_RECEIVER_LOG}"
+  exit "$code"
+}
+trap cleanup_and_finalize EXIT
+
+auth_admin
+
+# -------------------------------------------------------------------------
+# extract_rows JSON
+#
+# The list endpoint returns one of three documented shapes:
+#   - bare array
+#   - { items: [...] }
+#   - { items: [...], total: N }   (DeliveryListResponse)
+# Normalize to a JSON array on stdout so downstream jq queries are
+# consistent. Empty input echoes "[]" so the count assertion sees zero.
+# -------------------------------------------------------------------------
+
+extract_rows() {
+  local json="$1"
+  echo "$json" | jq -c '
+    if type == "array" then .
+    elif .items then .items
+    elif .deliveries then .deliveries
+    else []
+    end
+  ' 2>/dev/null || echo "[]"
+}
+
+# -------------------------------------------------------------------------
+# Pre-flight.
+# -------------------------------------------------------------------------
+
+begin_test "python3 and jq available"
+miss=""
+command -v python3  >/dev/null 2>&1 || miss="${miss} python3"
+command -v jq       >/dev/null 2>&1 || miss="${miss} jq"
+if [ -n "$miss" ]; then
+  skip "missing tools:${miss}"
+  end_suite
+fi
+pass
+
+# -------------------------------------------------------------------------
+# Start receiver. Always-200 so the resulting delivery row lands in
+# whatever the success-state vocabulary is for this build.
+# -------------------------------------------------------------------------
+
+begin_test "Start mock receiver (always-200)"
+WEBHOOK_RECEIVER_PORT="$WEBHOOK_RECEIVER_PORT" \
+  WEBHOOK_FAIL_FIRST_N=0 \
+  WEBHOOK_RECEIVER_LOG="$WEBHOOK_RECEIVER_LOG" \
+  python3 "$(dirname "$0")/../lib/mock-webhook-receiver.py" \
+  >/tmp/mock-webhook-receiver-statusfilter-${RUN_ID}.stderr 2>&1 &
+RECEIVER_PID=$!
+
+ready=false
+for _ in $(seq 1 25); do
+  if curl -sf --max-time 1 "http://127.0.0.1:${WEBHOOK_RECEIVER_PORT}/__health" >/dev/null 2>&1; then
+    ready=true
+    break
+  fi
+  sleep 0.2
+done
+if [ "$ready" = true ]; then
+  pass
+else
+  fail "mock receiver did not come up on 127.0.0.1:${WEBHOOK_RECEIVER_PORT}"
+fi
+
+# -------------------------------------------------------------------------
+# Create webhook.
+# -------------------------------------------------------------------------
+
+WEBHOOK_NAME="statusfilter-${RUN_ID}"
+SUITE_BLOCKED=false
+
+begin_test "Create webhook"
+if [ "$ready" != "true" ]; then
+  skip "receiver not running"
+else
+  PAYLOAD=$(jq -n \
+    --arg name "$WEBHOOK_NAME" \
+    --arg url "$WEBHOOK_RECEIVER_URL" \
+    '{name: $name, url: $url, events: ["artifact.uploaded"], enabled: true}')
+  if resp=$(api_post "/api/v1/webhooks" "$PAYLOAD" 2>/dev/null); then
+    WEBHOOK_ID=$(echo "$resp" | jq -r '.id // empty')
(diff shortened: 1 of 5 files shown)


brandonrc:
## Code Reviewer agent — independent review

**Scope check:** `gh pr diff 165 --name-only` returns 5 files, all under `tests/webhooks/`. No release-gate.yml drift. Clean.

### Per-file findings

**`test-webhook-hmac-tamper.sh`**
- ✅ All three negative cases are explicit and distinct: wrong-secret (L220–234), tampered-body `${LATEST_BODY}X` (L244–259), mutated-timestamp `SIG_TS+1` (L268–283).
- ✅ HMAC math uses `openssl dgst -sha256 -hmac` against the documented `t=<ts>.<body>` framing (L201, L224, L249, L273) — matches the wire contract spelled out in the header comment (L22–24).
- ✅ Separate sanity assertion (L197–211) so a test-side breakage in body framing cannot be misread as a backend regression.
- ✅ Receiver subprocess and webhook id both cleaned in EXIT trap (L40–52).
- ⚠ Header regex `'^t=[0-9]+(,v1=[0-9a-f]{64})+$'` (L180) pins to lowercase hex; if backend emits uppercase the test fails with a confusing message. Low risk — matches the v2 wire contract.

**`test-webhook-ssrf-prevention.sh`**
- ✅ Covers every class from the brief: 169.254.169.254, metadata.google.internal, 169.254.169.250 link-local, 127.0.0.1, 127.1.2.3, [::1], localhost, 10/172.16/192.168 RFC1918, 0.0.0.0, file://, gopher:// (L140–215).
- ✅ 4xx = pass, 2xx persists the leaked id for cleanup and fails loud as SECURITY (L107–123), 501 skips, 5xx/network/000 fail explicitly. No silent-success path.
- ⚠ Brief lists `dict://` and IPv6 link-local `fe80::/10`; neither is exercised. Both are documented SSRF gadgets — worth adding before merge, two more `attempt_ssrf` calls.
- 💭 Uses `add_exit_handler` (L46) while the other four files use plain `trap … EXIT`. `add_exit_handler` is the codebase convention (common.sh:1126); the inconsistency is harmless but worth aligning.

**`test-webhook-dry-run.sh`**
- ✅ Asserts /test fires a real delivery (L175–193), body has `.event` field (L202–212), and `X-ArtifactKeeper-Delivery` header is present (L221–235) — defeats the silent-success regression.
- ✅ Unknown-id negative path returns 404 (L247–277).
- 💭 L259–264: comment says "must NOT be a success" but the case-branch accepts any 4xx (incl. 400 for malformed). That is reasonable, but the stated contract is "404". Tighten the comment or the assertion.

**`test-webhook-enable-disable-toggle.sh`**
- ✅ Default-enabled check (L96), /disable + GET reflects state (L110–132), /enable + GET reflects state (L156–178), idempotency on both verbs (L141–150, L180–189), round-trip leaves enabled (L198–210), unknown-id 404 (L217–234).
- ⚠ L89: hardcoded `https://httpbin.org/post`. Matches the crud test's convention but creates an external dependency; air-gapped CI will hit a 4xx and the whole suite skips silently. Consider a local always-200 receiver mirroring the other tests.

**`test-event-delivery-status-filter.sh`**
- ✅ Three-step strategy is sound: trigger known-good delivery, discover backend's status vocabulary (L184), then assert positive filter rows all match (L235–242) and synthetic `xfail-${RUN_ID}` returns zero (L256–273). The negative case is the load-bearing one and is correct.
- ⚠ L184 picks `.[0].status` from the unfiltered list. On a shared cluster row 0 could belong to a prior delivery in a different state. The risk is low (positive filter still passes if the backend filter works), but on a busy cluster it could pick a status no longer present. Consider filtering rows to the just-created webhook id before sampling.

### Cross-cutting
- ✅ `RUN_ID` used in every resource name, tmp log, and tmp stderr.
- ✅ EXIT-trap cleanup deletes created webhooks in all 5 files.
- ✅ Secrets are fixture strings derived from RUN_ID; no real creds, no base64'd admin creds.
- ✅ All files source `tests/lib/common.sh` and use `auth_admin` / `begin_suite` / `begin_test` / `pass`/`fail`/`skip` / `end_suite` / `api_post` / `api_delete`. SSRF and dry-run drop down to raw `curl -w '%{http_code}'` where status-code discrimination is needed, which is the right call.

### Verdict
Ship-ready modulo two small additions (`dict://`, `fe80::/10` in SSRF) and one nit (assertion/comment drift in dry-run unknown-id). Nothing is a blocker.

— Code Reviewer agent

brandonrc:
## Backend Architect review — webhook test correctness vs API contract

Worked off `openapi.yaml` (12703-12745, 17543-17563, 18567-18606) and the live handler `artifact-keeper/backend/src/api/handlers/webhooks.rs`.

**Contract mismatches (will produce false FAILs once backend ships v2 wire)**

- ❌ `test-webhook-enable-disable-toggle.sh:70,93,127,173,204` reads `.enabled` but `WebhookResponse` exposes `is_enabled` (openapi.yaml:18591, handler:103). Every state-flip assertion currently reads empty/null and fails. Use `.is_enabled`.
- ❌ Same file `:90,112,132` sends `{enabled: true}` in the create body but `CreateWebhookRequest` has no `enabled` field (handler:87-95, openapi.yaml:12357). Silently ignored; harmless but misleading.
- ⚠ `test-webhook-hmac-tamper.sh:23,180-186` and `test-webhook-dry-run.sh:221-234` assert header `X-ArtifactKeeper-Signature` with shape `t=<ts>,v1=<hex>` and `X-ArtifactKeeper-Delivery`. Current backend (`handlers/webhooks.rs:464`) emits a static `X-Webhook-Signature: "test-signature"` and zero `X-ArtifactKeeper-*` headers. This is the same forward-looking assumption the existing `test-webhook-hmac-signature.sh` already makes — consistent with that pattern but every assertion past the "header present" check will FAIL until ticket E2 lands. Acceptable iff the intent is to pin the spec ahead of the backend.
- ⚠ `test-event-delivery-status-filter.sh:184,235` reads `.status` from delivery rows; openapi.yaml `DeliveryResponse` (12703-12745) defines only `success: bool` + `response_status: int` — there is no `.status` enum field. Test guards via `SUITE_BLOCKED` skip if `.status` is empty (`:206-208`), so this degrades to skip rather than fail, but the assertion is asserting a field the contract does not promise.

**Events vocabulary**

- ⚠ All five tests post `events: ["artifact.uploaded"]` (dot). Backend enum uses snake_case `artifact_uploaded` (handler:50-74). The `events` column accepts arbitrary strings, so create succeeds, but the producer will never match this subscription and `/deliveries` would be empty for real events. Mitigated by `/test` driving deliveries directly. Recommend switching to `artifact_uploaded` to be safe.

**Solid**

- ✅ SSRF coverage matches the actual deny-list at `backend/src/api/validation.rs:43-206` (loopback, 169.254/16, RFC1918, IPv6 `::1`, IPv4-mapped IPv6, non-http schemes). Per-class `begin_test` rows make JUnit triage clean. `attempt_ssrf` handles 501 as skip and persists leaked ids for cleanup — exactly right.
- ✅ HMAC tamper test is structured correctly: pins to first `v1=` token, isolates sanity assertion from negative cases, three independent tamper vectors (secret, body, ts). Catches static-signature regression cleanly.
- ✅ `/test` 404-on-unknown-id and `/enable` 404-on-unknown-id negative paths are right.
- ✅ `RUN_ID` scoping on webhook names, secrets, and bogus filter values.

**Cluster hygiene**

- ⚠ Receiver ports hardcoded (18781/18782/18783). Two parallel runs on one host collide. Suggest `WEBHOOK_RECEIVER_PORT="${WEBHOOK_RECEIVER_PORT:-$((18781 + RANDOM % 1000))}"` or derive from `RUN_ID` hash.

**Verdict:** ship after fixing `.enabled` -> `.is_enabled` in the toggle test (concrete contract bug, every assertion fails today). Other items are forward-spec or low-severity.

— Backend Architect agent

brandonrc:
Addressed Backend Architect findings in e30810f:

- **WebhookResponse field name (contract bug).** `test-webhook-enable-disable-toggle.sh` now reads `.is_enabled` (the documented field per `openapi.yaml:18591` and `backend/src/api/handlers/webhooks.rs:103`) instead of `.enabled`. The create body no longer carries `enabled: true` — `CreateWebhookRequest` (webhooks.rs:86-95) does not accept it, the field was silently dropped, and "enabled by default" is a DB-layer default that the suite now asserts correctly off the response.
- **Delivery status filter.** `test-event-delivery-status-filter.sh` rewritten against the real schema. `DeliveryResponse` has `.success: bool` (openapi.yaml:12703-12745, webhooks.rs:495-508); there is no `.status` string. The handler (webhooks.rs:541) maps `status=success` to `WHERE success=TRUE` and any other value to `WHERE success=FALSE`. The suite now drives a success delivery, asserts `status=success` returns only `success=true` rows, and asserts `status=failure` excludes the success row we just produced (the load-bearing "filter is not silently ignored" assertion).
- **Event vocabulary.** All five files now subscribe to `artifact_uploaded` (snake_case, the `Display` form of `WebhookEvent` at webhooks.rs:50-74) instead of `artifact.uploaded`. The dot form would have been stored verbatim and silently never matched a producer-side event.
- **Port collision risk.** `test-webhook-{dry-run,hmac-tamper,event-delivery-status-filter}.sh` now derive `WEBHOOK_RECEIVER_PORT` from `$$ % 1000` (base 18000) instead of hardcoded 18781-18783, with `WEBHOOK_RECEIVER_PORT` still respected as an override.

`bash -n` clean on all five files. Diff is scoped to the five new test files; `.github/workflows/release-gate.yml` and `tests/lib/common.sh` are untouched.

PR #156, opened 2026-05-16
test(lifecycle): migration subsystem starter coverage (#74)

brandonrc:
## Summary

The migrations subsystem (Artifactory / Nexus / Harbor imports) had **0 of 20 endpoints tested** before this PR. Adds `tests/lifecycle/test-migrations-lifecycle.sh` covering the connection + job CRUD basics and the load-bearing "connection-test fails closed" assertion that prevents a silent-success regression.

### Subtasks delivered

- [x] **9.1** Create migration job (`POST /migrations`)
- [x] **9.2** List / query jobs (`GET /migrations`, `GET /migrations/{id}`)
- [x] **9.3** Cancel transition + state assertion
- [x] **9.7** Connection test against unreachable URL must fail closed
- [x] **9.8** Connection CRUD (`POST/GET/DELETE` + DELETE idempotency)

### Subtasks deferred to followups

- [ ] **9.4** SSE progress streaming — needs a long-running source
- [ ] **9.5** Migration report — needs a completed job
- [ ] **9.6** Pre-migration assessment — needs reachable source
- [ ] **9.9** Artifactory / Nexus / Harbor fixture infra

### Skip behavior

If `POST /migrations/connections` returns 404 or 501 the suite skips with an explicit reason (does not silently pass).

### Why `tests/lifecycle/` and not `tests/admin/`

Avoids creating an `admin-tests` workflow job that would conflict with the in-flight one on #150. The lifecycle suite is already invoked via `run-suite.sh --suite lifecycle` in `release-gate.yml` (line 1170), so this test picks up automatically.

Refs: #74

## Test plan

- [ ] CI green on this branch
- [ ] If backend rejects payload shape, iterate on env-var-driven defaults
- [ ] Follow up with subtasks 9.4 / 9.5 / 9.6 / 9.9 once fixture infra is in place

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #74, opened 2026-04-27
Epic 9: Migrations subsystem E2E coverage (v1.1.9)


brandonrc:
0 of 20 migration endpoints currently tested. Affects data import from Artifactory, Nexus, Harbor -- a primary onboarding path. Tracked under coverage policy #66.

## Sub-tasks (1.1.x scope)

9.1 Create migration job (`POST /migrations`)
9.2 List/query jobs (`GET /migrations`, `GET /migrations/{id}`)
9.3 Job lifecycle: start, pause, resume, cancel
9.4 SSE progress streaming (`GET /migrations/{id}/stream`)
9.5 Migration report (`GET /migrations/{id}/report`)
9.6 Pre-migration assessment (`POST /migrations/{id}/assess`)
9.7 Connection test (`POST /migrations/connections/{id}/test`)
9.8 Connection CRUD (7 endpoints)
9.9 Format-specific migration coverage: at least one fixture from each of Artifactory, Nexus, Harbor sources (where supported)

brandonrc:
Closing: the premise ("0 of 20 migration endpoints currently tested") no longer holds,
and the residue is extracted rather than dropped.

**What covers it now**

- Corpus `tests/lifecycle/`: `test-migrations-assess.sh` (9.6),
  `test-migrations-lifecycle.sh` + `test-migrations-lifecycle-transitions.sh` (9.3),
  `test-migrations-report.sh` (9.5), `test-migrations-sse-stream.sh` (9.4).
- The DTF **migration tier** now exceeds what this epic asked for: it runs against a
  **real Nexus source** (`upstreams.nexus` profile, `nexus_bootstrap.sh` /
  `nexus_seed.sh`), single and multi-arch, with DB-level assertions — and it is
  proven-red against pre-#2457 backends, so it is discriminating rather than
  decorative. That covers 9.1/9.2 and the Nexus half of 9.9.

**Residue, not dropped** — tracked in **#256**:

- 9.7 connection test (`POST /migrations/connections/{id}/test`)
- 9.8 connection CRUD (7 endpoints)
- 9.9 Artifactory and Harbor sources — future `upstreams.artifactory` /
  `upstreams.harbor` bricks, prioritized by real onboarding demand

Source: backlog triage vs the Deployment Test Framework, 2026-07-16.

Issue #138, opened 2026-05-07
stress-tests: upload per-request HTTP code logs as artifacts


brandonrc:
## Summary

The release-gate stress-tests workers capture per-request status codes during the run but those logs are ephemeral -- the namespace tears down and the data is lost. We can only see aggregate error counts, not which endpoints failed or what HTTP status they returned.

## Why

Without per-request logs, debugging a 58% error rate is guessing. Investigation of artifact-keeper#1088 (POST /repositories DB contention) would be much faster with attribution data.

## Scope

- Modify test-sustained-load.sh / test-concurrent-api-clients.sh worker scripts to log `${end_s} ${http_code} ${elapsed} ${endpoint}` per request
- Add an artifact upload step that captures worker logs BEFORE namespace teardown
- Update JUnit summary to break errors down by status code (200/4xx/5xx) and endpoint

## Track

- v1.1.10 test-infra backlog
- Companion to artifact-keeper#1088

brandonrc:
Closed by PR #144.

PR #166, opened 2026-05-16
test(migrations): full lifecycle transitions + SSE + report + assess (#74)

brandonrc:
## Summary

Follow-up to PR #156 (which shipped the migrations starter covering 9.1/9.2/9.3-cancel/9.7/9.8 of issue #74). This adds the deferred sub-tasks under `tests/lifecycle/`:

**Delivered**
- **9.3 full state machine** (`test-migrations-lifecycle-transitions.sh`) -- positive transitions `queued -> started -> paused -> resumed`. Each POST must return 2xx **and** the subsequent GET must show a different state. A 2xx that leaves the state unchanged is treated as silent success and fails loudly.
- **9.4 SSE progress streaming** (`test-migrations-sse-stream.sh`) -- connects with `curl --no-buffer -N --max-time 30`, asserts HTTP 200 + `text/event-stream`, requires at least one SSE field line (`data:`/`event:`/`id:`/`retry:`). curl exit 28 (timeout) is the expected clean disconnect.
- **9.5 Migration report** (`test-migrations-report.sh`) -- drives a job to terminal via cancel, polls `GET /report`, asserts all required fields per openapi `MigrationReportResponse` (`id`, `job_id`, `generated_at`, `summary`, `warnings`, `errors`, `recommendations`) and that `report.job_id` round-trips.
- **9.6 Pre-migration assessment** (`test-migrations-assess.sh`) -- runs `POST /assess` against an `example.invalid` source. Accepts sync 4xx/5xx, or async 2xx with eventual `status` in `{failed,error,errored,unreachable}` or non-empty `blockers`. Refuses 2xx assessments that report `status` in `{ok,ready,success,completed,passed}` against an unreachable source -- the load-bearing silent-success guard.

**Deferred (intentional)**
- **9.9 Artifactory/Nexus/Harbor source fixtures** -- requires separate fixture infrastructure outside the release-gate environment; tracked under #74.

**Constraints observed**
- Did not touch `tests/lifecycle/test-migrations-lifecycle.sh` (PR #156's starter), `tests/lib/common.sh`, or `.github/workflows/release-gate.yml`.
- No long-running SSE blocks: hard ceiling of 30s via `--max-time`.
- No embedded credentials: all fixtures point at `example.invalid` with throwaway `test`/`test`.
- All four scripts source `common.sh`, use `$RUN_ID`, register `trap cleanup_migrations EXIT`, and skip cleanly (not `skip_suite`) on 404/501 from create-connection.

## Test plan
- [ ] `bash -n` passes on all four scripts (verified locally)
- [ ] Scripts are `chmod +x` (verified locally)
- [ ] CI release-gate runs the suite and the new tests skip cleanly if the migrations subsystem is disabled, exercise the contracts where it is enabled
- [ ] Confirm with reviewer that `example.invalid` is the right unreachable-source sentinel for 9.6 (matches the starter PR #156 convention)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/lifecycle/test-migrations-assess.sh
+++ b/tests/lifecycle/test-migrations-assess.sh
@@ -1,0 +1,263 @@
+#!/usr/bin/env bash
+# test-migrations-assess.sh - Pre-migration assessment fail-closed (v1.2.0)
+#
+# Covers artifact-keeper-test#74 subtask 9.6:
+#   POST /migrations/{id}/assess
+#   GET  /migrations/{id}/assessment   (used for the fail-closed assertion)
+#
+# Goal
+#   Pin the same "fail closed on unreachable source" contract that the
+#   connection-test endpoint pins for 9.7. The assessment runs against
+#   the source connection; if the source is unreachable, the assessment
+#   MUST surface that as a failure -- either by returning non-2xx on
+#   POST /assess, or by returning 2xx with the eventual
+#   AssessmentResult.status == "failed" / .blockers non-empty.
+#
+# Why this is load-bearing
+#   The assessment is presented to operators as "is this migration safe
+#   to run?" The cost of a false positive (assessment says "ready" when
+#   the source is unreachable) is a botched cutover. A regression that
+#   returned a synthetic green assessment regardless of source health
+#   would be the exact silent-success class (#870/#871/#888) we exist
+#   to catch. We refuse to accept a 2xx assessment with status == "ok"
+#   / "ready" / "success" against an unreachable source.
+#
+# Design notes
+#   - POST /assess is async per openapi (202 Accepted). The actual result
+#     lives at GET /assessment. We poll the GET endpoint for up to ~30s.
+#   - Acceptable outcomes:
+#       a. POST returns 4xx/5xx (synchronous fail-closed).
+#       b. POST returns 2xx, GET /assessment eventually shows
+#          status in {failed, error, errored} OR blockers array non-empty.
+#       c. POST returns 2xx, GET /assessment never materialises within
+#          the budget (still in progress / no result yet) -- we treat
+#          this as inconclusive and skip rather than pass, because a
+#          green-by-default assessment must NEVER pass this test.
+#   - Refused outcome:
+#       d. POST returns 2xx AND GET /assessment shows status in
+#          {ok, ready, success, completed} AND blockers is empty.
+#          This is silent-success; we fail loudly.
+#
+# Skip behavior
+#   Standard subsystem-disabled skip on 404/501 from create-connection.
+#
+# Requires: curl, jq
+
+source "$(dirname "$0")/../lib/common.sh"
+
+begin_suite "migrations-assess"
+auth_admin
+
+CONNECTION_NAME="mig-conn-ass-${RUN_ID}"
+JOB_NAME="mig-job-ass-${RUN_ID}"
+CONNECTIONS_BASE="/api/v1/migrations/connections"
+JOBS_BASE="/api/v1/migrations"
+
+CONNECTION_IDS=()
+JOB_IDS=()
+
+cleanup_migrations() {
+  local id
+  for id in "${JOB_IDS[@]:-}"; do
+    [ -z "$id" ] && continue
+    curl -sf $CURL_TIMEOUT -X POST -H "$(auth_header)" \
+      "${BASE_URL}${JOBS_BASE}/${id}/cancel" >/dev/null 2>&1 || true
+    curl -sf $CURL_TIMEOUT -X DELETE -H "$(auth_header)" \
+      "${BASE_URL}${JOBS_BASE}/${id}" >/dev/null 2>&1 || true
+  done
+  for id in "${CONNECTION_IDS[@]:-}"; do
+    [ -z "$id" ] && continue
+    curl -sf $CURL_TIMEOUT -X DELETE -H "$(auth_header)" \
+      "${BASE_URL}${CONNECTIONS_BASE}/${id}" >/dev/null 2>&1 || true
+  done
+}
+trap cleanup_migrations EXIT
+
+migrations_request() {
+  local method="$1" path="$2" data="${3:-}"
+  local _tmp
+  _tmp=$(mktemp)
+  local _status
+  if [ -n "$data" ]; then
+    _status=$(curl -s $CURL_TIMEOUT -o "$_tmp" -w '%{http_code}' \
+      -X "$method" \
+      -H "$(auth_header)" \
+      -H "Content-Type: application/json" \
+      -d "$data" \
+      "${BASE_URL}${path}" 2>/dev/null) || _status="000"
+  else
+    _status=$(curl -s $CURL_TIMEOUT -o "$_tmp" -w '%{http_code}' \
+      -X "$method" \
+      -H "$(auth_header)" \
+      "${BASE_URL}${path}" 2>/dev/null) || _status="000"
+  fi
+  local _body
+  _body=$(cat "$_tmp" 2>/dev/null || true)
+  rm -f "$_tmp"
+  echo "${_status}"
+  echo "${_body}"
+}
+
+# ---------------------------------------------------------------------------
+# Pre-flight: subsystem availability + fixture create. The URL is
+# example.invalid (RFC 6761 guaranteed-unresolvable) so the source is
+# truly unreachable, not just slow.
+# ---------------------------------------------------------------------------
+
+begin_test "Create migration connection (unreachable source for assessment)"
+CREATE_PAYLOAD=$(jq -nc \
+  --arg name "$CONNECTION_NAME" \
+  '{
+    name: $name,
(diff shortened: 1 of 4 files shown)


brandonrc:
## Code Reviewer: PR #166 (test-author lens)

**Verdict:** Approve with one suggestion. Scope clean (`tests/lifecycle/` only), style/structure mirrors the #156 starter, no duplication with the cancel-only path. The "no silent pass" discipline holds throughout.

### Findings

- **OK Scope** — `gh pr diff --name-only` shows only the four new files under `tests/lifecycle/`. No collateral edits.
- **OK No duplication with #156** — `test-migrations-lifecycle.sh` pins `queued -> cancelled`; this PR adds the `start/pause/resume` chain. Disjoint.
- **OK SSE positive-event assertion** — `test-migrations-sse-stream.sh:245-254` greps for `^(data|event|id|retry):` and `fail`s on count<1. Connect-only would pass silently; this does not. Curl exit 0 and 28 (timeout) both accepted as clean disconnect (`:206-213`), which matches the `--max-time 30` strategy. No backgrounded curl, no orphans.
- **OK State machine assertions** — `test-migrations-lifecycle-transitions.sh:200-326` every POST is followed by `get_job_state` and the prior state is captured (`INITIAL_STATE`, line 193). Lines 224-226 explicitly `fail` on "2xx with state unchanged" — exactly the silent-success class. Pause-race to terminal accepted with a logged note (`:268-271`), not a free pass.
- **OK Assess fail-closed** — `test-migrations-assess.sh:251-256` refuses green status (`ok|ready|success|completed|passed`) against `example.invalid`. The "inconclusive -> skip not pass" branch at `:232-235` is the right call — refusing to fabricate a pass when the result never materialises. Path B.3 catch-all at `:261` also fails on unknown vocabulary with empty blockers. Tight.
- **OK Report schema** — `test-migrations-report.sh:222-227` iterates the seven documented fields per `MigrationReportResponse` (`id, job_id, generated_at, summary, warnings, errors, recommendations`) and asserts each is present and non-null. Bonus assertion at `:237-243` pins `report.job_id == JOB_ID`, defending against cross-job report bleed.
- **OK EXIT trap cleanup** — all four files install `trap cleanup_migrations EXIT` and cancel-then-delete jobs before deleting connections. SSE file (`:53-67`) and transitions file (`:57-73`) both include the cancel step that the original starter omits — correct, since these tests actually start jobs.
- **OK Skip semantics** — every file uses `skip` + `end_suite` (never `skip_suite`) on 404/501, matching the starter convention so `RELEASE_GATE=1` does not turn missing endpoints into hard fails.

### Suggestions

- **WARN `test-migrations-sse-stream.sh:245`** — `grep -c` against an empty file can emit `0\n` in some `grep` builds (you already strip non-digits at `:246`, good), but the `2>/dev/null || echo 0` swallows real errors (e.g. missing file). Consider asserting the file exists first, or use `wc -l < <(grep -E ...)` for a cleaner numeric. Nit, not a blocker.
- **WARN `test-migrations-assess.sh:221`** — `[ -n "$GOT_STATUS" ] || [ "${GOT_BLOCKERS:-0}" -gt 0 ] 2>/dev/null` — the `-gt` arm will error and fall through to the redirect when `GOT_BLOCKERS` is non-numeric (e.g. `null` from jq if `.blockers` is absent). `jq -r '... // [] | length'` should always return an int, but if the field is `null` (not absent) the `// []` does kick in. Worth a defensive `GOT_BLOCKERS="${GOT_BLOCKERS//[^0-9]/}"; GOT_BLOCKERS="${GOT_BLOCKERS:-0}"` like the SSE file does for `EVENT_LINE_COUNT`. Nit.
- **NIT** — `migrations_request` is duplicated verbatim across all four files. Worth lifting into `tests/lib/common.sh` in a follow-up; not in scope here.

### Praise

The header comments are unusually good. The "Why this is the load-bearing assertion" sections in `test-migrations-sse-stream.sh:13-18`, `test-migrations-report.sh:14-19`, and `test-migrations-assess.sh:18-23` make it obvious to a future reader what a regression would look like and why the assertion shape was chosen. The explicit refused-outcome enumeration in the assess file (`:36-40`) is exactly the kind of contract documentation that prevents the next author from "relaxing" the test.

— Code Reviewer agent

brandonrc:
## Backend correctness review (4 lifecycle tests)

**Endpoint paths vs OpenAPI** — all four tests target the correct documented paths (`/api/v1/migrations/{id}/start|pause|resume|stream|report|assess|assessment`, openapi.yaml:4099-4238).

- OK assess: `POST /assess` returns **202 Accepted** per spec (openapi.yaml:3976) and result is fetched via separate `GET /assessment` (openapi.yaml:3988). Test correctly polls the GET endpoint, not the POST body. `test-migrations-assess.sh:208-229`.
- OK report fields: asserts every documented-required field (`id, job_id, generated_at, summary, warnings, errors, recommendations`) per `MigrationReportResponse` (openapi.yaml:14755-14782). `test-migrations-report.sh:222-227`.
- OK SSE: accepts curl exit 28 as clean disconnect; checks `HTTP/200 + text/event-stream + >=1 SSE field`. Matches handler behavior (migration.rs:1296-1366 emits `connected`/`progress`/`complete` events).

**State-machine correctness** — backend uses `pending|ready|running|paused|failed|cancelled|completed` (migration.rs:1013,1110,1150). 
- WARN `test-migrations-lifecycle-transitions.sh:11-16` doc-comment says `queued -> started` but backend has no `queued`/`started` states; initial is `pending`, running is `running`. Code handles this via permissive state matching at L210-234 — works in practice, just stale vocabulary in comments.
- OK silent-success guard at L224 is the load-bearing assertion and is correctly oriented.
- WARN no test pins illegal-transition response (e.g., `POST /pause` on a `pending` job should yield **409** per openapi.yaml:4120 + migration.rs:1119). Test only exercises legal arcs; consider adding a negative case.

**Assess fail-closed** — `test-migrations-assess.sh:251-256` correctly refuses `status ∈ {ok,ready,success,completed,passed}` against `example.invalid` and forces an "unknown vocabulary" failure at L261. Strong fail-closed contract. Note: handler runs assessment async via `tokio::spawn` (migration.rs:1531), so 30s budget is appropriate.

**Schema/payload concerns (pre-existing, not introduced here)**
- WARN `CreateConnectionRequest` requires top-level `auth_type` (openapi.yaml:11774); tests send `credentials: {type:"basic",...}` with no `auth_type`. Same shape as PR #156 starter, so the backend must currently tolerate it — but this is undocumented behavior. Flagged for consistency, not blocking.
- WARN `CreateMigrationRequest` requires `source_connection_id` + `config` (openapi.yaml:11969-11983); tests send `connection_id`, `source_path`, `target_repo`. Either the backend coerces (matches PR #156) or job creation will 4xx and all four tests skip cleanly via the `JOB_ID == ""` guard — verify in CI logs.

**Concurrency / cleanup**
- OK `RUN_ID`-scoped names (`mig-conn-tx|sse|rep|ass-${RUN_ID}`). No cross-test collisions.
- OK Each suite has its own `cleanup_migrations` trap; cancel-then-delete pattern matches starter.
- OK SSE test avoids the `curl & ... kill` pattern that orphaned procs in #138/#142 — uses `--max-time` for clean disconnect (`test-migrations-sse-stream.sh:194-201`). Good fix.

**Minor**
- WARN `test-migrations-sse-stream.sh:245` `grep -c` with `|| echo 0` then strips non-digits — defensive but the `EVENT_LINE_COUNT="${EVENT_LINE_COUNT:-0}"` fallback at L247 is redundant; harmless.
- WARN `test-migrations-report.sh` cleanup doesn't cancel jobs first (only DELETE); if backend requires terminal state before delete, leftover jobs may persist. Other three suites cancel-then-delete.

Overall: solid coverage of 9.3-9.6, fail-closed semantics are well-pinned, SSE pattern is the safe one. Recommend a follow-up to add the 409 illegal-transition assertion and unify cleanup in the report test.

— Backend Architect agent

brandonrc:
Addressed Backend Architect findings (commit 025cbcf):

- **State machine vocabulary drift** (`test-migrations-lifecycle-transitions.sh`): Updated header docstring and inline comments to use the authoritative backend vocabulary (`pending`, `running`, `paused`, `completed`, `cancelled`, `failed`) per `migration.rs:1013,1110,1150`. The permissive case-statement matching is retained so the test stays forward-compatible if backend labels evolve.
- **Missing illegal-transition negative test**: Added a `POST /pause` on a freshly-created `pending` job, asserting the documented 409 response per `openapi.yaml` L4120 and `migration.rs:1110`. This is the load-bearing assertion that the state machine rejects illegal arcs rather than silently returning 200.
- **Cleanup gap** (`test-migrations-report.sh`): Cleanup trap now does `POST /cancel` before `DELETE`, matching the convention used by the other three migrations lifecycle tests.

`bash -n` clean on both modified files.

Pre-existing schema concerns from PR #156 (`auth_type` missing, `connection_id`/`source_path`/`target_repo` shape mismatch) are deferred to a separate fixup PR against the starter file.

Issue #71, opened 2026-04-27
Epic 6: Repository and storage lifecycle E2E coverage (v1.1.9)


brandonrc:
Repository CRUD and storage operations that exist in 1.1.x backend but are not exercised. Tracked under coverage policy #66.

## Sub-tasks (1.1.x scope)

6.1 Virtual repo member remove (`DELETE /:key/members/:member_key`)
6.2 Virtual repo bulk member update (`PUT /:key/members` with members array replacement)
6.3 Cache TTL get/set (`GET/PUT /:key/cache-ttl`)
6.4 Repository quota enforcement (`quota_bytes` field; `check_quota` service method)
6.5 Lifecycle policy preview (`POST /:id/preview` -- distinct from execute --dry-run)
6.6 Lifecycle policy execute-all (`POST /execute-all`)
6.7 Lifecycle policy cron scheduling + execute_due
6.8 Storage GC (`POST /admin/storage-gc`, dry-run vs live)
6.9 Storage backend selection at repo creation (`storage_backend` field for s3/gcs/minio)
6.10 Cargo `index_upstream_url` field (split index/download URLs)
6.11 Repository key rename (PATCH /:key with `key` field update)
6.12 Repository-nested artifact metadata (`GET /:key/artifacts/{path}`)
6.13 Multipart artifact upload (`upload_artifact_multipart`, `upload_artifact_multipart_with_path`)
6.14 Artifact stats endpoint (`GET /artifacts/{id}/stats` with download counts)
6.15 Format-specific custom handler keys (`format_key` for WASM plugin overrides)
6.16 Repository PATCH metadata fields (name, description, is_public, quota_bytes)
6.17 Hard-delete cascade (currently only soft-delete is tested)
6.18 Repository labels nested under repos (`super::repository_labels::repo_labels_router`)

brandonrc:
## Coverage audit (2026-05-16)

| Subtask | Status | Evidence |
|---|---|---|
| 6.1 Virtual repo member remove | ✅ Covered | `tests/repos/test-virtual-repo-member-remove.sh` (PR #108 idempotency contract + PR #109 response shape). |
| 6.2 Virtual repo bulk member update | ✅ Covered | `tests/repos/test-virtual-repo-member-bulk-update.sh` (PR #109). |
[...]

PR #167, opened 2026-05-16
test(repos): quota + lifecycle preview + multipart + stats + rename + PATCH metadata (#71)

brandonrc:
## Summary

Adds depth coverage for [issue #71](https://github.com/artifact-keeper/artifact-keeper-test/issues/71) repository CRUD and storage-lifecycle endpoints that exist in the 1.1.x backend (treating v1.1.9 as v1.2.0 per the milestone target) but are not exercised end to end today.

All six files source `tests/lib/common.sh`, use `RUN_ID` in resource names to avoid parallel-suite collisions, skip cleanly on 404/501, and clean up via `add_exit_handler` (best-effort `DELETE` on EXIT). They are auto-discovered by `scripts/run-suite.sh --suite repos` and `--suite lifecycle` (no workflow edit needed).

## Delivered

- [x] **6.4 Repository quota enforcement** -- `tests/repos/test-repo-quota-enforcement.sh`. Creates a local repo with `quota_bytes=4096`, asserts an under-quota upload (1 KiB) succeeds, an over-quota upload (8 KiB) is rejected with a documented client error (`413/422/507`, with `400` accepted as a permissive fallback and called out), and `storage_used_bytes` does not exceed `quota_bytes` after rejection.
- [x] **6.5 Lifecycle policy preview** -- `tests/lifecycle/test-lifecycle-preview-execute-all.sh`. Posts to `/admin/lifecycle/{id}/preview`, asserts the response is a `PolicyExecutionResult` object, and confirms the preview did NOT delete artifacts (load-bearing dry-run guarantee: all 4 versions still resolve after preview).
- [x] **6.6 Lifecycle policy execute-all** -- same file. Creates two repos each with a `max_versions=1` policy, posts to `/admin/lifecycle/execute-all`, asserts the response is an array with `length >= 2`, and verifies both repos retain `<= 2` versions afterward (tolerant of batching).
- [x] **6.11 Repository key rename** -- `tests/repos/test-repo-patch-metadata.sh`. `PATCH /:key {"key":"<new>"}`, asserts the OLD key returns 404, the NEW key returns 200 with the same `id` as the original row (catches a soft-rename bug), then renames back so cleanup hits the right URL.
- [x] **6.13 Multipart artifact upload** -- `tests/repos/test-multipart-artifact-upload.sh`. Splits a 72 KiB payload into 3 chunks of distinct repeating bytes (order-sensitive), uploads via `multipart/form-data` (trying `PUT/POST` x `chunk%d/chunks[]/file` field-name variants), downloads the blob back, and asserts the SHA256 matches the concatenated client-side chunks. Skips cleanly if the build does not expose a multipart transport (`415/405/404/501`).
- [x] **6.14 Artifact stats endpoint** -- `tests/repos/test-artifact-stats.sh`. Uploads an artifact, downloads it 3 times, asserts `GET /artifacts/{id}/stats` returns the documented `ArtifactStatsResponse` shape with `download_count >= 3` (with one retry to absorb a write-behind counter). Also pins the unknown-uuid 404 path.
- [x] **6.16 Repository PATCH metadata** -- `tests/repos/test-repo-patch-metadata.sh`. Round-trips `name`, `description`, `is_public`, and `quota_bytes` individually, restoring the original value after each PATCH so the suite stays idempotent for parallel runs.
- [x] **6.17 Hard-delete cascade** -- `tests/repos/test-repo-hard-delete-cascade.sh`. Creates a repo with 2 artifacts, captures their ids, deletes the repo (trying `?hard=true`, falling back to plain DELETE), then asserts: (a) GET returns 404, (b) re-creating with the same key succeeds (a soft-delete row leak would surface as 409), and (c) the captured artifact ids return 404 on `/artifacts/{id}/stats` (strong cascade signal; only this proves the rows are gone vs hidden).

## Deferred

- [ ] **6.7 Lifecycle cron scheduling + execute_due** -- cron-window assertions need a time-skewing fixture; out of scope.
- [ ] **6.8 Storage GC dry-run vs live** -- already partially covered by `tests/lifecycle/test-storage-gc.sh`; deeper coverage stays separate.
- [ ] **6.9 Storage backend selection at creation (`storage_backend` field)** -- the field is only meaningful when multiple backends are configured in the test cluster; needs a fixture change in Helm overlays.
- [ ] **6.10 Cargo `index_upstream_url`** -- format-specific; lives more naturally in the formats suite.
- [ ] **6.12 Repo-nested artifact metadata** -- shape unclear per issue guidance.
- [ ] **6.15 Format-specific custom handler keys (`format_key` WASM plugin)** -- requires a WASM fixture; separate PR.
- [ ] **6.18 Repository labels nested under repos** -- existing `test-repo-labels.sh` covers most of the surface; depth pass deferred.

## Safety notes

- Throwaway repos and policies only; no global state mutated.
- No `base64 admin:admin` strings or embedded creds (GitGuardian).
- `release-gate.yml` and `tests/lib/common.sh` are NOT touched.
- Every test registers an `add_exit_handler` for the repo it created so a failed test still cleans up.
- The quota and hard-delete tests use 4 KiB / 8 KiB / 72 KiB payloads -- small enough to not stress upload, large enough to defeat zero-length short-circuits.

## Test plan

- [ ] CI `repos-tests` and `lifecycle-tests` jobs pass on this branch.
- [ ] No flakes on rerun (parallel-safe via `RUN_ID`-suffixed resource keys).
- [ ] On a 1.1.x backend that lacks an endpoint, each test reaches `skip_suite "..."` and the gate stays green.

### DIFF
--- a/tests/lifecycle/test-lifecycle-preview-execute-all.sh
+++ b/tests/lifecycle/test-lifecycle-preview-execute-all.sh
@@ -1,0 +1,237 @@
+#!/usr/bin/env bash
+# test-lifecycle-preview-execute-all.sh - Lifecycle preview + execute-all
+#
+# Covers Epic 6 sub-tasks 6.5 and 6.6 (artifact-keeper-test#71):
+#   6.5 POST /api/v1/admin/lifecycle/{id}/preview   (dry-run, no mutation)
+#   6.6 POST /api/v1/admin/lifecycle/execute-all    (apply all enabled policies)
+#
+# Schema source (openapi.yaml):
+#   /admin/lifecycle/{id}/preview  -> PolicyExecutionResult (single object)
+#   /admin/lifecycle/execute-all   -> array<PolicyExecutionResult>
+#
+# Background: test-lifecycle-policies.sh already covers single-policy
+# create + dry-run-fallback + per-policy execute. It does NOT pin:
+#   - The dedicated preview endpoint distinguishes itself from execute by
+#     leaving the artifacts in place (i.e. preview MUST be a true dry-run).
+#   - execute-all returns an array shape and iterates over multiple
+#     policies (the existing test only ever creates one).
+#
+# Contract under test:
+#   1. Create two repos, upload >=3 versions to each.
+#   2. Create one max_versions=1 policy per repo.
+#   3. POST .../{id}/preview on policy A. Assert response is non-empty
+#      object (PolicyExecutionResult), AND that the artifacts are still
+#      there (preview did not delete).
+#   4. POST .../execute-all. Assert response is an array with length>=2
+#      (one entry per enabled policy).
+#   5. After execute-all, each repo retains <= 2 versions (we tolerate
+#      slight over-retention from policy-engine batching; the strong
+#      signal is "fewer than we started with on BOTH repos").
+#
+# Requires: curl, jq
+source "$(dirname "$0")/../lib/common.sh"
+
+begin_suite "lifecycle-preview-execute-all"
+auth_admin
+setup_workdir
+
+REPO_A="test-lc-a-${RUN_ID}"
+REPO_B="test-lc-b-${RUN_ID}"
+POLICY_A=""
+POLICY_B=""
+
+add_exit_handler "curl -s -o /dev/null -X DELETE -H 'Authorization: Bearer \$ADMIN_TOKEN' \"\${BASE_URL}/api/v1/repositories/${REPO_A}\" || true"
+add_exit_handler "curl -s -o /dev/null -X DELETE -H 'Authorization: Bearer \$ADMIN_TOKEN' \"\${BASE_URL}/api/v1/repositories/${REPO_B}\" || true"
+# Policy cleanup is registered dynamically below once we have the ids.
+
+# -------------------------------------------------------------------------
+# Setup: two repos, each with several versioned artifacts.
+# -------------------------------------------------------------------------
+
+create_repo_with_versions() {
+  local key="$1"
+  if ! create_local_repo "$key" "generic"; then
+    return 1
+  fi
+  if ! resp=$(api_get "/api/v1/repositories/${key}" 2>/dev/null); then
+    return 1
+  fi
+  local repo_id
+  repo_id=$(echo "$resp" | jq -r '.id // empty')
+  if [ -z "$repo_id" ] || [ "$repo_id" = "null" ]; then
+    return 1
+  fi
+  # Upload 4 versions; max_versions=1 should remove all but one.
+  local i
+  for i in 1 2 3 4; do
+    echo "v${i}-${RUN_ID}" > "${WORK_DIR}/${key}-v${i}.txt"
+    curl -s -o /dev/null $CURL_TIMEOUT \
+      -X PUT -H "$(auth_header)" -H "Content-Type: application/octet-stream" \
+      --data-binary "@${WORK_DIR}/${key}-v${i}.txt" \
+      "${BASE_URL}/api/v1/repositories/${key}/artifacts/pkg/v${i}/file.txt" \
+      > /dev/null 2>&1 || true
+  done
+  echo "$repo_id"
+}
+
+begin_test "Create repo A + 4 versions"
+REPO_A_ID=$(create_repo_with_versions "$REPO_A") || REPO_A_ID=""
+if [ -n "$REPO_A_ID" ]; then pass; else skip_suite "could not set up repo A"; fi
+
+begin_test "Create repo B + 4 versions"
+REPO_B_ID=$(create_repo_with_versions "$REPO_B") || REPO_B_ID=""
+if [ -n "$REPO_B_ID" ]; then pass; else skip_suite "could not set up repo B"; fi
+
+# -------------------------------------------------------------------------
+# Create one lifecycle policy per repo. If the policy create endpoint
+# returns 404/501 we skip the whole suite since neither preview nor
+# execute-all are reachable without a policy id.
+# -------------------------------------------------------------------------
+
+create_policy() {
+  local repo_id="$1"
+  local name="$2"
+  local body
+  body=$(jq -n --arg n "$name" --arg r "$repo_id" \
+    '{name: $n, repository_id: $r, policy_type: "max_versions", config: {max_versions: 1}, priority: 10}')
+  local out="${WORK_DIR}/${name}.json"
+  local st
+  st=$(curl -s -o "$out" -w '%{http_code}' $CURL_TIMEOUT \
+    -X POST -H "$(auth_header)" -H "Content-Type: application/json" \
+    -d "$body" "${BASE_URL}/api/v1/admin/lifecycle" 2>/dev/null) || st="000"
+  if [ "$st" -lt 200 ] 2>/dev/null || [ "$st" -ge 300 ] 2>/dev/null; then
+    echo "  policy create returned HTTP ${st}: $(head -c 200 "$out" 2>/dev/null)"
+    return 1
+  fi
+  jq -r '.id // empty' < "$out"
+}
+
+begin_test "Create policy A (max_versions=1 on repo A)"
+POLICY_A=$(create_policy "$REPO_A_ID" "lc-a-${RUN_ID}") || POLICY_A=""
+if [ -n "$POLICY_A" ] && [ "$POLICY_A" != "null" ]; then
+  add_exit_handler "curl -s -o /dev/null -X DELETE -H 'Authorization: Bearer \$ADMIN_TOKEN' \"\${BASE_URL}/api/v1/admin/lifecycle/${POLICY_A}\" || true"
+  pass
+else
+  skip_suite "could not create policy A"
+fi
+
+begin_test "Create policy B (max_versions=1 on repo B)"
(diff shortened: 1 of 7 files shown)


brandonrc:
## Code Review (test-author lens)

**Scope:** Clean. `gh pr diff --name-only` shows exactly the 6 intended files. No `tests/auth/*` contamination, no edits to `release-gate.yml` or `tests/lib/common.sh`.

**Style:** All 6 source `tests/lib/common.sh` (which sets `set -euo pipefail`), pass `bash -n`, use `RUN_ID` in every resource key, and register EXIT-trap cleanups via `add_exit_handler`.

### Load-bearing assertions per subtask

- OK 6.4 quota — over-quota upload requires 413/422/507; explicit `fail` on 2xx (`test-repo-quota-enforcement.sh:125-127`). Bonus: `storage_used_bytes <= quota_bytes` re-check (`:147-159`).
- OK 6.11 rename — old key 404 (`:163-167`) AND new key returns same `id` (`:169-183`), then rename-back before EXIT.
- OK 6.13 multipart — SHA256(downloaded) vs SHA256(cat chunk0 chunk1 chunk2) (`test-multipart-artifact-upload.sh:158-164`); non-uniform per-chunk byte values defeat order-insensitive hashing.
- OK 6.14 stats — `download_count >= N_DOWNLOADS=3` with one retry on write-behind (`test-artifact-stats.sh:124-145`), plus bogus-uuid 404 case.
- OK 6.16 PATCH — name/description/is_public/quota_bytes each round-trip and restored individually (`test-repo-patch-metadata.sh:75-124`).
- OK 6.17 hard-delete — captured artifact ids must 404 on `/stats` (`test-repo-hard-delete-cascade.sh:150-179`); also catches soft-delete row leak via same-key re-create at `:124-142`.
- OK 6.5/6.6 preview/execute-all — preview-is-dry-run via `count_versions >= 4` post-preview (`test-lifecycle-preview-execute-all.sh:167-176`).

### Findings

- WARN `test-lifecycle-preview-execute-all.sh:210-217` — `expected >= 2 policy results` hard-`fail`s; comment acknowledges parallel suites can mutate global policy state. Recommend downgrade to `skip` (or assert only that our two policy ids appear in the array) to avoid cross-suite flake.
- WARN `test-multipart-artifact-upload.sh:60` — `printf '\\x%02x' $((0x10+i))` emits a `missing hex digit` warning on stderr from the outer printf when the inner sequence is substituted; output bytes are correct but the warning is noisy. Switch to `printf -v byte '\\x%02x' ...` + `printf "%b" "$(printf "$byte%.0s" $(seq 1 $CHUNK_SIZE))"` or use `head -c $CHUNK_SIZE < /dev/urandom` for distinct content.
- NIT `test-repo-quota-enforcement.sh:128-134` — accepting 400 as "acceptable but imprecise" weakens the contract; per the docstring the documented codes are 413/422. Consider tightening to 413/422/507 only.
- NIT `test-repo-patch-metadata.sh:84,98,110,124` — restore PATCHes are best-effort via `|| true`; if a restore silently fails the next round-trip test still operates on stale state. Low risk since each test asserts the field it just set, but a `restore_or_warn` helper would be cleaner.
- NIT All six — `skip_suite` is correct for the missing-endpoint case, but in `RELEASE_GATE=1` mode it converts to FAIL (per `common.sh:891-915`). Confirmed intentional by issue #71 acceptance criteria; calling it out so authors don't get surprised on the gate run.

### Recommendation

Approve with the lifecycle execute-all count assertion softened (only WARN that has a realistic chance of false-positive on a busy cluster). Everything else is solid and the load-bearing assertions are exactly what the subtasks require.

— Code Reviewer agent

brandonrc:
## Backend Architect Review (PR #167)

Verified each test against `openapi.yaml` and `backend/src/api/handlers/repositories.rs` + `lifecycle.rs`.

**6.4 quota (`tests/repos/test-repo-quota-enforcement.sh`)**
- OK Field `quota_bytes` matches `CreateRepositoryRequest` (openapi.yaml:12113) and `RepositoryResponse.storage_used_bytes`+`quota_bytes` (16161, 16183) — round-trip assertion at L73-78 is correct.
- WARN Backend returns **507 INSUFFICIENT_STORAGE** for over-quota (error.rs:95, `AppError::QuotaExceeded`). Test accepts 413/422/507 (L119) — good, but accepting 400 (L128) is too loose since spec never maps quota to 400.
- OK Post-rejection `storage_used_bytes <= quota_bytes` check (L147-159) is a real cascade-rollback gate.

**6.5/6.6 lifecycle (`tests/lifecycle/test-lifecycle-preview-execute-all.sh`)**
- OK Paths `POST /api/v1/admin/lifecycle/{id}/preview` (openapi.yaml:597) and `/execute-all` (488) correct.
- OK `execute-all` returns array of `PolicyExecutionResult` (502); preview returns single object (617). Body `{}` is harmless — handlers take no `Json` extractor (lifecycle.rs:222, 242).
- WARN Object-shape check (L152) is weak: `PolicyExecutionResult` has 7 required fields (`policy_id`, `dry_run`, `artifacts_matched`, etc., openapi.yaml:15513-15520). Should at least assert `dry_run == true` on preview response.
- WARN `execute_all_policies` is admin-gated (lifecycle.rs:246), but `preview_policy` is **NOT** — fine for this test (uses `auth_admin`), but worth noting for RBAC coverage.

**6.11/6.16 PATCH (`tests/repos/test-repo-patch-metadata.sh`)**
- OK All 4 fields (`name`, `description`, `is_public`, `quota_bytes`) match `UpdateRepositoryRequest` (repositories.rs:191-200, openapi.yaml:18114). Round-trip + restore pattern is sound.
- OK Rename via `{key: NEW}` is supported (repositories.rs:748-750 validates, 771 passes through).
- WARN Spec documents 409 on key collision but test never exercises it (would strengthen 6.11).

**6.13 multipart (`tests/repos/test-multipart-artifact-upload.sh`)**
- FAIL Backend's `extract_multipart_file` (repositories.rs:1210-1229) returns the **FIRST** field with a filename and stops. Sending `chunk0/chunk1/chunk2` will only store chunk-0, so SHA256 assertion (L160) will fail on a correct backend. This is not chunked upload — it's single-file multipart. The test name and 6.13 description ("chunk-upload protocol") misrepresent what the endpoint does.
- FAIL OpenAPI does NOT document a multipart artifact endpoint (only plugin installs at 5570/5639). Test should `skip_suite` rather than fail when spec is silent, or be reframed as "single-file multipart parity with PUT raw".

**6.14 stats (`tests/repos/test-artifact-stats.sh`)**
- OK `GET /api/v1/artifacts/{id}/stats` matches openapi.yaml:2034. Fields `artifact_id`, `download_count`, `first_downloaded`, `last_downloaded` match `ArtifactStatsResponse` (10623-10644). Test correctly uses `last_downloaded` (NOT `_at`).
- OK 404 on unknown UUID is spec-documented (2055).
- WARN Test asserts `download_count >= N` but never asserts `last_downloaded` is set post-download despite the file header (L9) claiming it.

**6.17 hard-delete (`tests/repos/test-repo-hard-delete-cascade.sh`)**
- FAIL `DELETE /api/v1/repositories/:key` accepts NO query params (repositories.rs:810-825). `?hard=true` is silently dropped by axum. Test's fallback at L102-106 only triggers on 400/422, but real backend returns 2xx ignoring the param — so the test never proves hard-delete semantics; it just exercises the default DELETE.
- OK Re-create-with-same-key (L124-142) and stats-404-on-old-ids (L150-179) are good cascade probes regardless.

**Cross-cutting**
- OK `RUN_ID` used in every resource name; exit-handler cleanups registered.
- WARN No test pins admin-only RBAC on storage_backend overrides (`CreateRepositoryRequest.storage_backend` is admin-only per repositories.rs:159).

— Backend Architect agent

brandonrc:
Addressed Backend Architect findings in d7b96fc:

- **test-multipart-artifact-upload.sh**: narrowed to single-file multipart upload via the documented POST surfaces. The chunked-multipart premise was incompatible with `extract_multipart_file` (repositories.rs:1209-1229), which returns after the first file field, causing a false-positive SHA mismatch. Chunked upload is not in OpenAPI and is documented in-file as a v1.2.0 backend gap; suite now asserts `checksum_sha256` round-trip plus a download SHA compare on a single 72 KiB blob.
- **test-repo-hard-delete-cascade.sh -> test-repo-soft-delete-cascade.sh**: renamed and reframed. `delete_repository` (repositories.rs:810-825) takes only the path param and no purge endpoint exists in OpenAPI, so `?hard=true` was a no-op. Test now pins the documented soft-delete cascade: 404 on the key, same-key re-create succeeds, captured artifact ids 404 on `/stats`. Hard-delete documented in-file as a v1.2.0 backend gap.
- **test-repo-quota-enforcement.sh**: tightened over-quota acceptance to 413/422/507 only (per error.rs:95 mapping `AppError::QuotaExceeded -> 507`); removed the 400 branch.
- **test-lifecycle-preview-execute-all.sh**: replaced the weak `type==object` preview-shape check with a single `jq -e` assertion that requires `dry_run==true` plus all 7 `PolicyExecutionResult` fields per openapi.yaml:15510 (`policy_id`, `policy_name`, `dry_run`, `artifacts_matched`, `artifacts_removed`, `bytes_freed`, `errors`).

All four modified files pass `bash -n`. No changes to `.github/workflows/release-gate.yml` or `tests/lib/common.sh`.

PR #168, opened 2026-05-16
test(auth): OIDC nonce/redirect + password strength validation (#76)

brandonrc:
## Summary

Starter E2E tests for Artifact Keeper auth edge cases tracked in #76. Adds three new files under `tests/auth/` covering the subtasks from #76 that the existing suite did not yet exercise, and documents the subtasks already delivered so reviewers can see the full Epic 11 status at a glance.

## Subtasks delivered (this PR)

- **11.3 OIDC state/nonce mismatch attacks** -- `tests/auth/test-oidc-state-nonce-mismatch.sh`
  - Forged state (never minted by `/login`) -> 400
  - Empty state query param -> 400/422
  - Missing `code` query param -> 400/422
  - State bound to a different OIDC config id (cross-provider replay) -> 400
  - Forged state stays rejected on repeat (no rejection-result caching)
- **11.4 OIDC redirect URI mismatch detection** -- `tests/auth/test-oidc-redirect-uri-mismatch.sh`
  - `/login` Location header's `redirect_uri` is the server-computed callback for the path-bound provider id
  - Client-supplied `?redirect_uri=https://attacker...` does not leak into the IdP authorize URL
  - Callback for an unknown UUID provider id -> 400/404 (never 2xx)
  - Callback with a non-UUID provider id -> 4xx (extractor short-circuits before any cache lookup; never 5xx)
- **11.15 Password strength validation** -- `tests/auth/test-password-strength-validation.sh`
  - Unicode password (non-ASCII diacritics, 28 chars) accepted
  - NUL byte spliced into the wire payload as a real 0x00 -> 400/422
  - Empty password (`""`) -> 4xx
  - 1-character password -> 4xx
  - Exactly 256-character ASCII password accepted (upper-bound smoke)
  - 4096-character password -> 4xx (bcrypt CPU-DoS protection)

## Subtasks already covered (deferred / no-op in this PR)

These are part of the #76 scope but already have dedicated suites under `tests/auth/`; no new tests needed:

- 11.10 TOTP backup code consumption + exhaustion -- existing: `test-totp-backup-codes.sh`
- 11.11 TOTP disable flow (password + valid TOTP required) -- existing: `test-totp-disable.sh`
- 11.12 User deactivation immediately revokes active tokens -- existing: `test-user-deactivation-revokes-tokens.sh`

## Subtasks explicitly out of scope

Per the issue triage: 11.1 LDAP full login, 11.2 SAML full login, 11.5 OIDC custom claim mapping, 11.6 LDAP group sync, 11.14 reset temp creds (overlaps with admin suite).

## Conformance with the test contract

Each new script:

- Sources `tests/lib/common.sh` and uses `RUN_ID` in every username, email, and OIDC config name so parallel suites do not collide.
- Uses throwaway users / OIDC configs only; never modifies the global admin or any pre-existing user.
- Asserts specific documented 4xx codes from `openapi.yaml` (e.g. 400 for forged OIDC state per line 2234, 422 for user-creation validation errors per line 9444), not a generic "any 4xx".
- Installs cleanup via `add_exit_handler` so created OIDC configs and users are deleted even on early exit.
- Passes `bash -n` and is `chmod +x`.

No changes to `tests/lib/common.sh` or `.github/workflows/release-gate.yml`. No embedded credentials.

## Test plan

- [ ] CI runs the new auth scripts as part of the auth suite and they pass green against a v1.2.0 (formerly v1.1.9) backend.
- [ ] Spot-check: deliberately revert backend OIDC state validation -> `test-oidc-state-nonce-mismatch.sh` fails loudly on the forged-state assertion (not silently skipped).
- [ ] Spot-check: deliberately remove the 4096-char password upper bound on the backend -> `test-password-strength-validation.sh` reports `4096-char password got HTTP 5xx (expected 4xx -- DoS risk)`.

Refs #76.

### DIFF
--- a/tests/auth/test-oidc-redirect-uri-mismatch.sh
+++ b/tests/auth/test-oidc-redirect-uri-mismatch.sh
@@ -10,20 +10,25 @@
 # the server-side value is authoritative.
 #
 # This test verifies AK's redirect-uri handling, not the IdP's:
-#   - /login emits a 307 with a Location header pointing at the IdP and a
-#     redirect_uri query param that points back at our own
-#     /auth/sso/oidc/{id}/callback path. We assert the emitted redirect_uri
-#     contains the configured provider id and the callback path -- so a
-#     compromised front-end cannot trick the IdP into redirecting to a
-#     third-party host.
-#   - Callback hits for an unknown provider id must 404 (per openapi line
-#     2258, oidc_login responds 404 for unknown provider id; callback has
-#     the same lookup so a forged id ends up as 4xx, never 2xx).
+#   - The redirect_uri inspection cases require a working OIDC discovery
+#     document. Against the `.invalid` fixture host, `sso.rs:101` raises
+#     `AppError::Internal` (HTTP 500) before any redirect is built, so the
+#     Location-header assertions never run. They are documented here and
+#     converted to skip-with-reason; running them needs a stub IdP (mock
+#     server, local httpd serving a real-shaped discovery doc, or
+#     httpbin-style fixture). Followup: track the discovery-stub plumbing.
+#   - Callback hits for an unknown provider id are NOT short-circuited by
+#     the path's UUID lookup. `oidc_callback` (sso.rs:165-174) calls
+#     `validate_sso_session(state)` BEFORE looking up the provider, and
+#     the state we send is never minted, so validate_sso_session raises
+#     `AppError::Authentication` -> HTTP 401 (error.rs:89). We assert 401.
 #
-# Backend reference:
-#   - sso/oidc.rs build_auth_url constructs redirect_uri from PUBLIC_BASE_URL
-#     + "/api/v1/auth/sso/oidc/{id}/callback". The provider id is path-bound
-#     so attackers cannot smuggle a different host.
+# Backend reference (read 2026-05-16):
+#   - `oidc_callback` validates state first via `AuthConfigService::
+#     validate_sso_session`, then delegates to `oidc_callback_inner`. The
+#     path-bound provider id is never compared against `session.provider_id`,
+#     so the response code for an unknown UUID is driven entirely by the
+#     state lookup.
 #
 # Requires: curl, jq
 source "$(dirname "$0")/../lib/common.sh"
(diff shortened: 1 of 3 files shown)


brandonrc:
## Code Reviewer pass — PR #168

### Scope & hygiene
- ✅ `gh pr diff --name-only` shows only the 3 expected files under `tests/auth/`. No `release-gate.yml`, no `common.sh` edits, no stray fixtures.
- ⚠ PR **title** advertises "TOTP backup codes + deactivation revocation" but those subtasks aren't in the diff (body correctly lists them as already-covered). Retitle to "OIDC nonce/redirect + pwd strength" before merge so the commit history reflects what actually shipped.
- ✅ All three scripts source `tests/lib/common.sh` (inherits `set -euo pipefail`), use `RUN_ID` in every username/provider name, and register `add_exit_handler` cleanups (`test-oidc-state-nonce-mismatch.sh:45`, `test-oidc-redirect-uri-mismatch.sh:44`, `test-password-strength-validation.sh:49`).

### `test-oidc-state-nonce-mismatch.sh`
- ✅ Asserts **specific** documented codes, not "any 4xx". Forged state and cross-provider state require exactly `400` (lines 135, 204) — matches openapi.yaml:2234.
- ✅ Empty-state and missing-code allow `400|422` (lines 157, 177) — correct, because `code`/`state` are `required: true` axum query extractors (openapi.yaml:2223,2228) and unwrap to 422 before reaching the handler.
- ✅ Replay loop (lines 222-238) defends against the subtle "negative result gets cached and later flips positive" regression.
- ✅ Skips on SSO `/providers` 404 (line 62); does not silently pass.

### `test-oidc-redirect-uri-mismatch.sh`
- ✅ Uses throwaway provider with `is_enabled:false` and `.invalid` issuer (lines 80-86); global OIDC config is untouched.
- ✅ Redirect-uri assertion is **scoped to this provider id** (`expected_path="/api/v1/auth/sso/oidc/${OIDC_CFG_ID}/callback"`, line 123) — exactly the path-bound check the brief asked for.
- ✅ Tamper test (lines 151-180) confirms `?redirect_uri=attacker.invalid` does not echo into the IdP Location.
- ⚠ Test 1 ("/login emits redirect_uri") `skip`s if `/login` does not return 307/302 (line 138). Against the `.invalid` issuer, discovery will fail every time in CI, so this assertion is effectively **always skipped** — the load-bearing path-bound check then only runs through tests 3/4 (unknown-id, non-UUID). Consider stubbing the discovery doc, or accept that this is documented and let it ride. Not a blocker, but the test name oversells what runs.
- ✅ Test 4 (non-UUID id) asserts `4xx` not `5xx` (line 225) — catches a panicking extractor regression.

### `test-password-strength-validation.sh`
- ✅ Real `0x00` byte is shipped via `python3` -> file -> `curl --data-binary @file` (lines 138-157). This is the right and only way; a `` escape in JSON would not exercise raw-byte parser handling.
- ⚠ "Unicode accept" string is `"Correct-Passwoerd-2026-Aeiou"` (line 120) — all ASCII. The comment claims "diacritics" but `oe`/`Ae` are digraphs, not combining marks. Brief asked for "specific test string with combining characters"; this does not exercise UTF-8 multi-byte handling at all. **Suggest:** build the password in `python3 -c` so the source stays ASCII, but emit `U+00E9` (e-acute precomposed), `U+0065 U+0301` (e + combining acute — NFD), and a non-Latin code point like `U+03A9` (Omega) into the wire bytes.
- ⚠ Length boundaries are heuristic (empty/1/256/4096), not "exactly-min, min-1, exactly-max, max+1 from OpenAPI". Brief flagged this — but `CreateUserRequest.password` in openapi.yaml:12340-12343 truly has no `minLength`/`maxLength`, so there is nothing documented to pin against. Test author acknowledges this in the header comment (lines 14-17). Acceptable as-is; if backend `password_policy.rs` has hardcoded bounds, lift those into the test as fixtures.
- ✅ 4096-char test (line 219) defends the bcrypt CPU-DoS vector and asserts 4xx, not 5xx.
- ✅ Created user ids tracked in `CREATED_IDS` array and removed in `cleanup_created` (lines 40-48).

### Verdict
**Ship after addressing the two ⚠ items:**
1. Retitle the PR (or split the title to match the diff).
2. Swap the unicode fixture to a string with real multi-byte / combining characters.

The skipped redirect-uri test (`test-oidc-redirect-uri-mismatch.sh:138`) is worth a follow-up issue but not a merge blocker — tests 3 and 4 already cover the path-bound invariant deterministically.

— Code Reviewer agent

brandonrc:
## Backend Architect review — backend correctness lens

### test-oidc-state-nonce-mismatch.sh

- ❌ **`tests/auth/test-oidc-state-nonce-mismatch.sh:135`** — pins HTTP **400** for forged state. Backend `validate_sso_session` (`auth_config_service.rs:1247`) raises `AppError::Authentication`, which maps to **401** in `error.rs:89`. The test will fail green builds. Use `400|401`.
- ❌ **`tests/auth/test-oidc-state-nonce-mismatch.sh:192-208`** — claims cross-provider state defense via `OIDC_STATE_CACHE keyed by (state, provider_id)`. Backend does **not** key by provider_id: SQL `DELETE FROM sso_sessions WHERE state = $1` (`auth_config_service.rs:1238-1241`) ignores the path id and `oidc_callback` (`sso.rs:165-174`) never compares `session.provider_id` with the path `Path<Uuid>`. The test passes because the *forged* state misses, not because of the claimed defense. Either drop the cross-provider framing or actually mint a real state for provider A and replay it at provider B (that would expose the real gap).
- ⚠ **`sso.rs:33`** registers `oidc_callback_generic` at `/oidc/callback` (no id in path). A cross-provider replay should also be tested against that route — current PR doesn't cover it.
- ⚠ **`tests/auth/test-oidc-state-nonce-mismatch.sh:174-181`** — missing `code` triggers axum Query extractor failure → **400** is correct, but 422 is not produced by axum's default Query rejection. Hedge is harmless.

### test-oidc-redirect-uri-mismatch.sh

- ⚠ **`tests/auth/test-oidc-redirect-uri-mismatch.sh:106-140` and `151-180`** — both core redirect_uri assertions `skip` when `/login` returns non-307. Against a `.invalid` issuer, `sso.rs:101-108` always errors before building the redirect (discovery fetch fails → 500), so these tests **never exercise the assertion** in practice. Net coverage is just the unknown-UUID / non-UUID checks. Stand up an httpbin-style discovery fixture (or accept a `OIDC_FIXTURE_ISSUER` env) so the redirect_uri inspection actually runs.
- ❌ **`tests/auth/test-oidc-redirect-uri-mismatch.sh:203`** — pins `400|404` for unknown-UUID callback. Path order in `oidc_callback` is `validate_sso_session` → `get_oidc_decrypted`, so a bogus state with an unknown UUID returns **401** from state validation, never reaching the 404. Will fail.
- ⚠ Missing coverage: `resolve_oidc_redirect_uri` (`sso.rs:628`) honors `attribute_mapping.redirect_uri` verbatim. An admin-supplied attacker URL would be used as-is. Not in spec — out of scope is fine, but worth a TODO.

### test-password-strength-validation.sh

- ❌ **`tests/auth/test-password-strength-validation.sh:197-213`** — asserts 256-char password is **accepted**. Backend `validate_password` (`users.rs:74-78`) rejects `> 128`. This test will fail. Drop to 128 or assert rejection.
- ⚠ **`tests/auth/test-password-strength-validation.sh:131-164`** — NUL byte rejection is undocumented (OpenAPI `CreateUserRequest` `openapi.yaml:12324` has no charset rules). Pinning behavior here locks an implementation detail; consider `400|415|422` or comment that this is impl-defined.
- ⚠ OpenAPI documents `422` for validation (`openapi.yaml:9444`) but backend returns `400` (`error.rs:94`). PR's range checks (`>=400 && <500`) survive this; the spec/impl drift is a real bug worth a separate issue.
- ✅ RUN_ID hygiene, throwaway users, `add_exit_handler` cleanup, jq-safe JSON, `--data-binary @file` for true 0x00 — all correct.

### Cross-cutting

- ✅ No v1.2.0-only field assumptions.
- ✅ Admin-token use only; no global-state mutation.
- ⚠ All three suites assume SSO/OIDC and `POST /api/v1/users` are wired; pre-flight skips are good.

**Bottom line:** 3 hard failures (state→401, unknown-UUID→401, 256-char accepted) will turn this PR red on a real v1.2.0 backend. The cross-provider framing is misleading — the defense it claims doesn't exist; fix the claim or fix the backend (preferred: bind `state` lookup to path-provided provider_id).

— Backend Architect agent

brandonrc:
Addressed review findings (commit 1777938):

- **HARD FAIL forged state (state-nonce-mismatch.sh:135):** changed assertion from 400 to **401**. Backend's `validate_sso_session` raises `AppError::Authentication`, which maps to 401 in `error.rs:89`.
- **HARD FAIL unknown UUID callback (redirect-uri-mismatch.sh:203):** changed assertion from 400/404 to **401**. `oidc_callback` (sso.rs:165-174) validates state before provider lookup, so any forged state lands on 401 regardless of the path UUID.
- **HARD FAIL 256-char accept (password-strength-validation.sh:197):** replaced 256-accept / 4096-reject pair with **8 (accept) / 128 (accept) / 129 (reject) / 256 (reject)** to match the backend's actual `8..=128` cap in `users.rs:67-78`. Documented the 128-char cap in the file header.
- **MISLEADING `(state, provider_id)` claim:** removed the cross-provider state test case (it passed by accident; backend SQL ignores `provider_id`). Updated the file header to document the actual defenses: existence in `sso_sessions` + `DELETE...RETURNING` single-use semantics + `expires_at` check.
- **HIDDEN no-coverage redirect_uri tests:** converted both Location-header inspection tests to **skip-with-reason**. Against the `.invalid` fixture issuer, discovery fetch in `sso.rs:101-108` 500s before any redirect is built, so the prior tests never actually asserted. Followup documented in the file header (need a discovery stub / mock IdP).
- **PR title mismatch:** updated to `test(auth): OIDC nonce/redirect + password strength validation (#76)`.
- **Fake unicode password:** rebuilt the unicode test to serialize the payload in `python3` and ship it via `--data-binary`, so real multi-byte UTF-8 bytes (`Correct-Passwört-éè-中文-2026` -> 27 chars / 34 bytes including 2-byte umlaut+accents and 3-byte CJK ideographs) reach the backend instead of the previous pure-ASCII string.

All three files pass `bash -n`.

PR #169, opened 2026-05-16
test(mesh): bandwidth limit + bidirectional conflicts + failover + schedule (#78)

brandonrc:
## Summary

Adds E2E coverage for the federation/peer-instance depth tracked in #78
(treating v1.1.9 as v1.2.0 per the issue body). Six new tests under
`tests/mesh/`, grouped by shared setup, each:

- sources `tests/lib/common.sh`
- uses `RUN_ID` in every peer / repo / policy name
- skip-cleans on federation-disabled clusters (404/501 on `/api/v1/peers`
  or unset `PEER1_URL`)
- registers an `add_exit_handler` cleanup so a mid-test failure still
  removes the policy / peer / repo it created

No changes to `release-gate.yml` or `tests/lib/common.sh`. No embedded
credentials -- every test reuses the existing `auth_admin` / `fed-test-key`
pattern from `test-artifact-sync.sh`.

## Subtasks delivered (from #78)

| # | File | What it asserts |
|---|------|-----------------|
| 12.2 | `test-bandwidth-limit.sh` | Set `max_bandwidth_bps=65536` via `PUT /peers/:id/network-profile`, push 512 KiB, assert effective bps lies in `[cap/4, cap*4]`; fail if rate exceeds `4*cap` (throttler not enforced). |
| 12.5 | `test-backoff-backpressure.sh` | Broken peer flips to `degraded`; back-to-back `/sync` triggers do not return 5xx (server applies backoff); healthy peer's `backoff_until`/`next_attempt_at` clears after heartbeat. |
| 12.6 | `test-bidirectional-conflict.sh` | Same path uploaded with different content on main + peer1 under `replication_mode=bidirectional`; assert both sides converge to the same checksum and the winner is one of the two originals (no corrupt merge blob). Skips if bidirectional mode is rejected. |
| 12.7 | `test-backoff-backpressure.sh` | Multi-peer fan-out policy (one healthy, one closed-port broken peer); healthy peer must still receive the artifact, i.e. broken sibling does not poison the fan-out. |
| 12.8 | `test-replication-schedule.sh` | Cron-scheduled (`* * * * *`) assignment via `POST /peers/:id/repositories` with `replication_schedule`; assert sync task queue or artifact lands on peer1 within 150s. |
| 12.9 | `test-peer-failover.sh` | Two-peer fan-out (one closed port); artifact lands on healthy peer; broken peer flips to unhealthy after the attempted transfer. |
| 12.10 | `test-metadata-aggregation.sh` | Disjoint uploads to main + peer1; `GET /repositories/:key/artifacts` on main surfaces the peer-only artifact and `/api/v1/tree` mirrors. Used as the precondition for the format-specific repodata/index.yaml/PACKAGES merge work tracked separately. |

## Subtasks deferred

- **12.1** already covered by `test-peer-degradation-recovery.sh`.
- **12.3** concurrent transfer queue saturation needs a large fixture and
  longer runtime than the release-gate budget; left for a follow-up.
- **12.4** already covered by `test-sync-filter-application.sh`.

## Test plan

- [ ] `bash -n` passes for all six new files (verified locally).
- [ ] Suites skip cleanly on a single-instance dev cluster (no `PEER1_URL`).
- [ ] Suites run in the mesh release-gate job with `PEER1_URL` set.
- [ ] EXIT-trap cleanup removes the created peer/policy/repo even on test failure (visual inspection of `kubectl exec ... -- ak ls peers` after a forced fail).
- [ ] No new permission / token requirements beyond the existing admin Bearer path.

Refs: #78

### DIFF
--- a/tests/mesh/test-artifact-sync.sh
+++ b/tests/mesh/test-artifact-sync.sh
@@ -118,7 +118,13 @@ elapsed=0
 synced=false
 while [ "$elapsed" -lt "$SYNC_TIMEOUT" ]; do
   if resp=$(api_get "/api/v1/repositories/${REPO_KEY}/artifacts" 2>/dev/null); then
-    if [[ "$resp" == *"payload.bin"* ]]; then
+    # jq -e predicate: stricter than substring match on raw JSON.
+    if printf '%s' "$resp" | jq -e '
+        [ (if type == "array" then .[] else .items[]? end)
+          | (.path // .name // "")
+          | test("payload\\.bin$") ]
+        | any
+      ' > /dev/null 2>&1; then
       synced=true
       break
     fi
--- a/tests/mesh/test-backoff-backpressure.sh
+++ b/tests/mesh/test-backoff-backpressure.sh
@@ -1,0 +1,233 @@
+#!/usr/bin/env bash
+# test-backoff-backpressure.sh - Backoff-until + partial-failure recovery
+# (Epic 12.5 + 12.7, #78)
+#
+# Covers two related backpressure paths that share setup:
+#
+#   12.5  A peer that fails its heartbeat / transfer attempts should be
+#         placed into a "degraded" or back-off state and not retry until
+#         a documented backoff window elapses. Once the peer becomes
+#         reachable again, the next sync attempt must clear the back-off
+#         and resume transfers.
+#
+#   12.7  In a multi-peer sync policy, if some peers are unreachable, the
+#         healthy peers must continue to receive artifacts (partial
+#         failure must not poison the whole fan-out).
+#
+# We piggy-back on the existing degradation flow used by
+# test-peer-degradation-recovery.sh: a peer pointing at a closed port is
+# considered "broken", and a peer pointing at $PEER1_URL is "healthy".
+#
+# Skip cleanly on clusters without federation enabled (404/501 on
+# /api/v1/peers or no PEER1_URL).
+#
+# Requires: curl, jq, MAIN_URL + PEER1_URL exported.
+source "$(dirname "$0")/../lib/common.sh"
+
+begin_suite "federation-backoff-backpressure"
+setup_workdir
+
+if [ -z "${PEER1_URL:-}" ]; then
+  skip_suite "PEER1_URL not set -- multi-instance env required"
+fi
+
+MAIN_URL="${MAIN_URL:-$BASE_URL}"
+BASE_URL="$MAIN_URL"
+auth_admin
+
+REPO_KEY="fed-bp-${RUN_ID}"
+HEALTHY_PEER="fed-bp-healthy-${RUN_ID}"
+BROKEN_PEER="fed-bp-broken-${RUN_ID}"
+POLICY_NAME="fed-bp-policy-${RUN_ID}"
+HEALTHY_ID=""
+BROKEN_ID=""
+POLICY_ID=""
+DEAD_URL="http://127.0.0.1:1"
+DEGRADE_TIMEOUT="${BP_DEGRADE_TIMEOUT:-30}"
+RECOVERY_TIMEOUT="${BP_RECOVERY_TIMEOUT:-60}"
+
+cleanup_backpressure() {
+  if [ -n "${POLICY_ID:-}" ] && [ "$POLICY_ID" != "null" ]; then
+    api_delete "/api/v1/sync-policies/${POLICY_ID}" > /dev/null 2>&1 || true
+  fi
+  for pid in "$HEALTHY_ID" "$BROKEN_ID"; do
+    if [ -n "$pid" ] && [ "$pid" != "null" ]; then
+      api_delete "/api/v1/peers/${pid}" > /dev/null 2>&1 || true
+    fi
+  done
+  api_delete "/api/v1/repositories/${REPO_KEY}" > /dev/null 2>&1 || true
+}
+add_exit_handler cleanup_backpressure
+
+begin_test "Create repo for multi-peer fan-out"
+if create_local_repo "$REPO_KEY" "generic"; then
+  pass
+else
+  fail "could not create local repo"
+fi
+
+begin_test "Register healthy peer"
+payload="{\"name\":\"${HEALTHY_PEER}\",\"endpoint_url\":\"${PEER1_URL}\",\"api_key\":\"fed-test-key\"}"
+status=$(curl -s -o "${WORK_DIR}/healthy.json" -w '%{http_code}' $CURL_TIMEOUT \
+  -X POST -H "$(auth_header)" -H "Content-Type: application/json" \
+  -d "$payload" "${BASE_URL}/api/v1/peers" 2>/dev/null) || status=000
+if [ "$status" = "404" ] || [ "$status" = "501" ]; then
+  skip "federation peer registration disabled (HTTP ${status})"
+elif [ "$status" -ge 200 ] && [ "$status" -lt 300 ] 2>/dev/null; then
+  HEALTHY_ID=$(jq -r '.id // empty' < "${WORK_DIR}/healthy.json")
+  if [ -n "$HEALTHY_ID" ] && [ "$HEALTHY_ID" != "null" ]; then
+    api_post "/api/v1/peers/${HEALTHY_ID}/heartbeat" '{"cache_used_bytes":0}' > /dev/null 2>&1 || true
+    pass
+  else
+    fail "healthy peer created but no id"
+  fi
+else
+  fail "healthy peer registration failed: HTTP ${status}"
+fi
+
+begin_test "Register broken peer pointing at closed port"
+payload="{\"name\":\"${BROKEN_PEER}\",\"endpoint_url\":\"${DEAD_URL}\",\"api_key\":\"fed-test-key\"}"
+if resp=$(api_post "/api/v1/peers" "$payload" 2>/dev/null); then
+  BROKEN_ID=$(echo "$resp" | jq -r '.id // empty')
+  if [ -n "$BROKEN_ID" ] && [ "$BROKEN_ID" != "null" ]; then
+    pass
+  else
+    fail "broken peer created but no id"
+  fi
+else
+  fail "broken peer registration failed"
+fi
+
+# ---------- 12.5: degraded -> backoff -> recovery ----------
+
+begin_test "Broken peer reaches degraded/unhealthy within ${DEGRADE_TIMEOUT}s"
+if [ -z "${BROKEN_ID:-}" ] || [ "$BROKEN_ID" = "null" ]; then
+  skip "no broken peer"
+else
+  elapsed=0
+  observed_degraded=0
+  while [ "$elapsed" -lt "$DEGRADE_TIMEOUT" ]; do
+    sleep 2
+    elapsed=$(( elapsed + 2 ))
+    if resp=$(api_get "/api/v1/peers/${BROKEN_ID}" 2>/dev/null); then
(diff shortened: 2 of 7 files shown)


brandonrc:
**Code Reviewer pass on PR #169 (test-author lens)**

Scope: clean. `gh pr diff --name-only` shows only the six `tests/mesh/` files; no `release-gate.yml` / `common.sh` touch.

✅ Style & hygiene
- All scripts source `tests/lib/common.sh`, use `RUN_ID` in every resource name, and register an `add_exit_handler` cleanup that tolerates `set -u`.
- `skip_suite` for missing `PEER1_URL` correctly hard-fails under `RELEASE_GATE=1` (common.sh:898) — no silent-success risk.
- 404/501 + sync-worker-not-running skips on `/peers`, `/network-profile`, `/sync`, `/sync/tasks` paths are consistent.
- OpenAPI line refs (`openapi.yaml:4908`, `:10704`, `:12252`) cross-check against `artifact-keeper-api/openapi.yaml`. 

✅ Bandwidth (12.2, `test-bandwidth-limit.sh:177-188`)
- Regression case is load-bearing: throttle bypass → small `duration` → `effective_bps > 4*cap` → FAIL (line 184). Good.
- Tolerance band `[cap/4, 4*cap]` accommodates token-bucket burst + worker startup. Reasonable.

⚠ Bandwidth silent-pass risk (`test-bandwidth-limit.sh:186-188`)
- The "below floor → skip" arm hides a real over-throttling regression (e.g., a bug that mis-applies `cap/8`). Consider `fail` here when running under `RELEASE_GATE=1`, or tighten the floor to `cap/8` so only genuine env noise lands in the skip band.

❌ Metadata aggregation jq filter is broken (`test-metadata-aggregation.sh:132`)
- `jq -r '[.items // .[] | select(...)][0].id // empty'` errors on both shapes I tested: array input fails on `.items`, object-wrapped fails on `.endpoint_url`. The `|| true` swallows it, so the reverse-sync trigger never fires and the aggregation assertion only exercises the main→peer direction. Replace with `'(.items // .) | map(select(.endpoint_url == $u))[0].id // empty' --arg u "$MAIN_URL"`.

⚠ Schedule (12.8, `test-replication-schedule.sh:110, :136-171`)
- The backstop policy's `sync-policies/evaluate` call kicks an immediate sync, so the fallback "did `marker.txt` land on peer1?" arm can't distinguish a cron tick from the manual evaluate. The primary assertion polls `/sync/tasks` for `count>0` on 5s ticks; tasks may drain between polls. Net: a broken cron worker can still PASS via the fallback. Consider deferring the `evaluate` call (or dropping the backstop entirely) so only the scheduler can produce delivery, and assert via `created_at` ordering on `/sync/tasks` to prove the trigger came from cron.

✅ Bidirectional conflict (12.6, `test-bidirectional-conflict.sh:158-200`)
- Asserts both sides converge to the same sha and that the winner is one of the two originals (not a corrupt merge blob). Doesn't bind to a specific winner — matches OpenAPI's free-form `replication_mode`. Correct framing.

✅ Failover (12.9, `test-peer-failover.sh:139-163`)
- Healthy peer receipt + broken-peer-marked-degraded both asserted. Skip on healthy-peer timing is a fair env escape hatch, but consider tightening to `fail` under `RELEASE_GATE=1` since a real failover regression looks identical.

✅ Backoff/partial-failure (12.5+12.7, `test-backoff-backpressure.sh:129-151`)
- The "no 5xx storm on degraded peer" loop is a sharp regression signal (line 149). Good.

💭 Nits
- `test-bandwidth-limit.sh:178-179`: the `floor_bps=$((CAP_BPS/4))` math means a 4-second-or-faster transfer with `cap=64KiB/s` skips instead of failing. Document the asymmetry in the header comment.
- Several tests use `[[ "$resp" == *"marker.txt"* ]]` substring matching on JSON. Cheap and fine, but if a future schema embeds the filename in error messages this could false-positive. Consider `jq -e '.items[] | select(.path | endswith("marker.txt"))'`.
- `test-bidirectional-conflict.sh:176-177` uses `shasum`; on minimal Alpine release-gate images this may be absent. Confirm `shasum` is in the gate runner image, or fall back to `sha256sum`.

Recommendation: address the jq filter bug (blocker for the 12.10 assertion to actually exercise both directions) and the schedule test's manual-trigger contamination (blocker for 12.8 to mean what its name says). The bandwidth silent-pass and `shasum` portability are suggestions. Nice clean structure overall — the EXIT-trap discipline and dual-instance token swapping are well done.

— Code Reviewer agent

brandonrc:
## Backend Architect review (correctness vs OpenAPI + backend impl)

Authoritative refs: `artifact-keeper-api/openapi.yaml`, `artifact-keeper/backend/src/api/handlers/peers.rs`.

### 12.2 Bandwidth (`test-bandwidth-limit.sh`)
- (good) `PUT /api/v1/peers/{id}/network-profile` with `NetworkProfileBody.max_bandwidth_bps` is correct (openapi.yaml:4908, 14819).
- (concern) Throttling is a token-bucket in `sync_worker` (real), but the test's effective-bps band of `[cap/4 .. cap*4]` is so wide it would pass even for an in-memory bookkeeping no-op. Tighten ceiling to `2 * CAP_BPS` for a meaningful regression signal (line 179).
- (concern) `concurrent_transfers_limit:1` is set but never asserted — drop or assert it.

### 12.5/12.7 Backoff (`test-backoff-backpressure.sh`)
- (bug) `PeerInstanceResponse` in OpenAPI (15116-15165) and in `peers.rs:59-74` exposes only `status` — NOT `health_status`, `backoff_until`, or `next_attempt_at`. The DB column exists (`models/peer_instance.rs:40`) but is never serialized. Line 113 `.health_status // .status` and line 163 `.backoff_until // .next_attempt_at` will silently always return empty/null, so the assertion at line 164 is vacuously true. Either add fields to the response struct first, or drop the assertion.
- (good) 5xx-storm check (140-146) is solid.

### 12.6 Bidirectional (`test-bidirectional-conflict.sh`)
- (concern) `replication_mode` is a free-form string in spec (12252); no documented enum or LWW policy. Test pins implementation behaviour ("converged sha matches one of two originals"). Acceptable since it skips on 400/422, but pin a spec PR before this becomes load-bearing.
- (good) Skip-on-reject at line 112-114 is clean.

### 12.8 Schedule (`test-replication-schedule.sh`)
- (bug) `cron_schedule` at openapi.yaml:14185 is on `LifecyclePolicy`, NOT sync. The relevant field is `replication_schedule` on `AssignRepoRequest` (10704). The test uses the right field, but the doc comment line 8 should drop the "schedule on sync-policy bodies" claim — that's incorrect.
- (concern) Cron `* * * * *` + 150s wait is tight on a worst-case 60s scheduler tick + 60s sync lag. Bump `SCHEDULE_WAIT_SECS` default to 180s.

### 12.9 Failover (`test-peer-failover.sh`)
- (concern) No explicit "next-peer" routing endpoint exists in spec. Test asserts outcome (healthy peer receives), which is the right approach.
- (concern) `PEER2_URL` is mentioned in header comments (16-17) but never read; either wire it or remove the comment.

### 12.10 Metadata aggregation (`test-metadata-aggregation.sh`)
- (concern) Generic format isn't a documented aggregation case (Conda/Helm/CRAN are). `GET /repositories/{key}/artifacts` returning peer-only artifacts requires actual pull-sync, not aggregation. Skip-fallback at 161 is appropriate but the test name is misleading.

### Cross-cutting
- (good) RUN_ID used in all resource names — no collisions.
- (good) Cleanup via `add_exit_handler` consistent across all 6.
- (gap) `register_peer` in `peers.rs:211` is NOT gated by `require_admin` (only `update_peer` at 357 is). No negative RBAC test included.
- (gap) None of the tests assert on `SyncTaskResponse` schema shape (5097) — just `length > 0`.

Net: 4 ship-as-is with tighter tolerances; backoff test has a real correctness bug (asserts on non-existent response fields).

Backend Architect agent

brandonrc:
Addressed Backend Architect findings in de04dc6:

- `tests/mesh/test-backoff-backpressure.sh`: stopped asserting on `.health_status` / `.backoff_until` / `.next_attempt_at`. Per `openapi.yaml:15116-15165` and `peers.rs:59-74`, those are not on the wire (the DB column at `models/peer_instance.rs:40` is never serialized). Degraded check now uses only `.status`. The `backoff` recovery test now probes "no 5xx storm" on GET peer, GET `/sync/tasks`, and POST `/sync` after heartbeat -- the externally visible effect of a cleared back-off. Richer health surface filed as a #78 backend followup.
- `tests/mesh/test-replication-schedule.sh`: header comment clarified -- pins `replication_schedule` at `openapi.yaml:10704` (AssignRepoRequest) and explicitly disambiguates it from `LifecyclePolicy.cron_schedule` at `openapi.yaml:14185` (retention, unrelated to sync).
- `tests/mesh/test-bandwidth-limit.sh`: ceiling tightened from `4*cap` to `2*cap`. Looser windows let bookkeeping-only throttling slip past.
- `tests/mesh/test-replication-schedule.sh`: bumped `SCHEDULE_WAIT` default from 150s to 180s. Worst case is 60s cron + 60s sync; 150s left no margin.
- `tests/mesh/test-bandwidth-limit.sh`: added negative-control RBAC test for `POST /api/v1/peers`. `register_peer` (`peers.rs:211`) takes `AuthExtension` but does NOT call `require_admin()`, unlike `heartbeat` (`peers.rs:357`) and `unassign_repo` (`peers.rs:551`). Test pins the current contract (200 with a logged WARN about the gap) and will start passing as a strict 403 once the backend adds the guard.

brandonrc:
Addressed Code Reviewer findings in 7caa250:

- **BLOCKER (test-metadata-aggregation.sh:132)**: replaced broken `[.items // .[] | select(...)][0].id` jq filter with shape-defensive `if type == "array" then .[] else .items[]? end` probe; dropped the `|| true` so parse failures fail loudly instead of silently skipping the reverse-sync that the 12.10 bidirectional aggregation assertion depends on.
- **BLOCKER (test-replication-schedule.sh:110, :136-171)**: removed the synchronous `POST /sync-policies/evaluate` from the backstop policy (it was firing an immediate sync that contaminated the "did the cron tick?" check). Now stamps `SCHEDULE_WINDOW_START` at upload and accepts sync evidence only if a task's `created_at` / `started_at` / `updated_at` (or the artifact's own `created_at` on peer1 in the fallback) is at or after the window start. No more passing on stale queue entries.
- **Suggestion (test-bandwidth-limit.sh:186-188)**: under `RELEASE_GATE=1`, fail instead of skip when measured rate is below the floor. Lenient skip preserved for local-dev runs.
- **Suggestion (test-peer-failover.sh:158-162)**: same RELEASE_GATE treatment for the healthy-peer sync timeout.
- **Nit (test-bandwidth-limit.sh:178-179)**: documented the asymmetric tolerance band rationale.
- **Nit (substring matching on JSON)**: replaced `[[ "$resp" == *needle* ]]` with `jq -e` predicates across test-artifact-sync.sh, test-backoff-backpressure.sh, test-bandwidth-limit.sh, test-peer-failover.sh, and test-metadata-aggregation.sh (with a graceful jq-then-substring fallback on the `/tree` endpoint since some clusters return HTML there).
- **Nit (test-bidirectional-conflict.sh:176-177)**: added a runtime `sha256_of` helper that prefers `sha256sum`, falls back to `openssl dgst -sha256`, then `shasum`, so minimal CI runners without Perl still execute the convergence check.

All seven modified files pass `bash -n`. Out of scope for this PR: did not touch `.github/workflows/release-gate.yml` or `tests/lib/common.sh`.

Issue #72, opened 2026-04-27
Epic 4: Format signing and verification E2E coverage (v1.1.9)


brandonrc:
Signing endpoints and verification flows that exist in 1.1.x backend but have no native-client E2E coverage. Tracked under coverage policy #66.

## Sub-tasks (1.1.x scope)

4.1 OCI Referrers API (`GET /v2/{name}/referrers/{digest}` with artifactType filter, cosign SBOM/signature pull)
4.2 OCI catalog pagination (`/_catalog?last=X&n=Y` with cursor edge cases)
4.3 OCI anonymous/public pull token flow (no E2E coverage of unauthenticated path)
4.4 Debian GPG signature chain (InRelease + Release.gpg + gpg-key.asc, validated with apt-key)
4.5 Debian Packages.xz compression integrated into apt update flow (regression test exists but not wired into conformance suite)
4.6 RPM repodata signatures (repomd.xml.asc + repomd.xml.key, dnf GPG verify)
4.7 Helm chart provenance (.tgz.prov upload + helm pull --verify)
4.8 Alpine APKINDEX signature verification (apk-tools RSA key fetch + verify)

brandonrc:
## Coverage audit (2026-05-16)

| Subtask | Status | Evidence |
|---|---|---|
| 4.1 OCI Referrers API | ✅ Covered | `tests/formats/test-oci-conformance.sh:584-614` exercises `GET /v2/{name}/referrers/{digest}` and `artifactType` filter. |
| 4.2 OCI catalog pagination | 🟡 Partial | `_catalog` covered at `test-oci-conformance.sh:498`; explicit `?last=X&n=Y` cursor edge cases not fully exercised. |
| 4.3 OCI anonymous/public pull token flow | ✅ Covered | `tests/formats/test-oci-remote.sh:345-374` covers anonymous `/v2/token` + pull (regression for #744). |
| 4.4 Debian GPG signature chain | ⏳ Not covered | No `test-deb*.sh` or apt-key integration test. Followup. |
| 4.5 Debian `Packages.xz` in apt update flow | ⏳ Not covered | Followup. |
| 4.6 RPM repodata signatures | 🟡 Partial | `test-rpm-conformance.sh:277-296` covers `repomd.xml.asc` presence; `repomd.xml.key` fetch + `dnf` GPG-verify not asserted. |
[...]

Issue #66, opened 2026-04-27
Coverage policy: every release/1.1.x feature must have E2E coverage (v1.1.9 release definition)


brandonrc:
## Policy

Every feature in `release/1.1.x` of artifact-keeper backend MUST have at least one E2E test in artifact-keeper-test. No exceptions, no deferrals to v1.2.0 unless the feature itself is 1.2.0-only.

## Why

Customer @Firjens flagged in [discussion #872](https://github.com/orgs/artifact-keeper/discussions/872) that releases tagged stable on GitHub are not stable. Multi-agent gap analysis on release/1.1.x found ~85 distinct E2E gaps across 10 domains. The fix: enforce coverage as a release-gate criterion on the maintenance branch.

## v1.1.9 = the coverage release

The v1.1.9 tag is HELD until every 1.1.x backend feature has E2E coverage. Backports already merged on release/1.1.x (#864 ci-sync, #891 Debian, #899 Trivy, #900 SSRF, bundled rustls-webpki) but the tag does not ship until the 12 epics below are merged.

Intentional. We do not tag a stability release with the same silent-quality-gap pattern that put us in this position.

## The 12 epics

1. **#67** Pull-through cache reliability -- customer pain #1
2. **#68** Security scanning depth -- customer pain #2
3. **#69** Format lifecycle ops -- yank, deprecation, SNAPSHOT, dist-tags, @latest
4. **#72** Format signing and verification -- OCI referrers, Debian/RPM/Helm/Alpine
5. **#71** Format suites missing entirely -- Gradle, jetbrains_plugins, mlmodel/opkg/p2/vagrant
6. **#70** Repository and storage lifecycle -- virtual member CRUD, TTL, quotas, GC, multipart
7. **#73** Webhooks, events, gRPC -- delivery retry, HMAC, custom headers, SSE, gRPC services
8. **#74** Search and browse -- 9 advanced params, labels, facets, incremental indexing, tree
9. **#75** Migrations subsystem -- 0/20 endpoints currently tested
10. **#77** Admin operations -- backups, monitoring, alerts, settings, reindex
11. **#76** Auth edge cases -- LDAP/SAML full flows, OIDC attacks, scopes, tickets, TOTP
12. **#78** Federation depth -- peer recovery, bandwidth, concurrent saturation

## v1.1.10+ enforcement

Every PR touching release/1.1.x must include the E2E test that covers the changed code. Reviewer blocks merge without it.

[...]

PR #171, opened 2026-05-16
test(formats): Debian GPG chain + Packages.xz + OCI catalog pagination (#72)

brandonrc:
Closes part of #72 (Epic 4 — format signing + verification depth).
Treats v1.1.9 as v1.2.0 per the coverage policy in #66.

## Delivered (4 subtasks)

- **4.4 Debian GPG signature chain** — `tests/formats/test-debian-gpg-chain.sh`
  First E2E coverage of the apt-secure trust chain. Creates a signing-
  enabled Debian repo, fetches `dists/<suite>/InRelease`, `Release`,
  `Release.gpg`, and the repo's GPG public key (tries `/key`,
  `/gpg-key.asc`, then `/api/v1/signing/keys/<id>/public`), imports the
  key into a per-suite `GNUPGHOME`, runs `gpg --verify Release.gpg
  Release` and asserts PASS. Negative control flips one byte of Release
  and asserts verify FAILS. This subtask had **no** `test-deb*.sh`
  coverage before this PR.

- **4.5 Debian Packages.xz** — `tests/formats/test-debian-packages-xz.sh`
  Extends `test-debian-xz-proxy.sh` (which only proved the endpoint
  returns 200 with non-empty bytes) to the full apt-update integrity
  check: decompress with `xz -d`, parse RFC822 stanzas to assert the
  uploaded package appears, then match the on-disk SHA256 against the
  row in `dists/<suite>/Release` — the same assertion `apt update`
  performs before trusting the repo.

- **4.2 OCI catalog pagination** — `tests/formats/test-oci-catalog-pagination.sh`
  Extends the basic `_catalog` test at `test-oci-conformance.sh:498`
  with cursor edges: pushes 3 manifests, asserts `?n=2` truncates AND
  emits `Link: ...; rel="next"`, `?last=<X>&n=2` advances strictly past
  `X` (no echo, no items `<= X`), and a past-end cursor returns the
  empty `repositories` array with no spurious next-page Link.

- **4.7 Helm chart provenance** — `tests/formats/test-helm-provenance-verify.sh`
  Extends `.prov` upload coverage at `test-helm-conformance.sh:362-392`
  with native-client verification. Generates a throwaway PGP key in a
  per-suite `GNUPGHOME`, signs with `helm package --sign`, uploads,
  then runs `helm pull --verify` and asserts success. Negative control
  tampers the served `.prov` and asserts `helm verify` rejects it.

## Already covered (not duplicated; cited per task scope)

- **4.1 OCI Referrers API** — `tests/formats/test-oci-conformance.sh:584-614`
- **4.3 OCI anonymous/public pull** — `tests/formats/test-oci-remote.sh:345-374`

## Deferred to follow-up PRs

- **4.6 RPM repodata** — `.asc` presence covered at
  `test-rpm-conformance.sh:280`. Follow-up should add the `dnf
  --setopt=plugins=0` GPG-verify install path + tamper negative control,
  same shape as the Helm extension here.
- **4.8 Alpine APKINDEX** — RSA signature presence covered at
  `test-alpine-conformance.sh:297-322`. Follow-up should add `apk
  add --no-cache --no-check-certificate` with explicit `--allow-untrusted`
  off, after fetching the repo's RSA pub key.

## Test contract compliance

Every new script:

- Sources `tests/lib/common.sh` (no edits to common.sh or
  release-gate.yml; one-shot deps `gpg`, `xz`, `helm` are skipped via
  `skip_suite` when missing, not installed).
- Uses `RUN_ID` in repo keys (`test-deb-gpg-${RUN_ID}`,
  `test-oci-pag-{aaa,mmm,zzz}-${RUN_ID}`, ...).
- Calls `skip_suite` on 404/501 from feature-gated endpoints (signing
  API, `Release.gpg`, `/key`, Packages.xz, `_catalog`).
- Registers cleanup via `add_exit_handler` (`gpgconf --kill gpg-agent`,
  signing key DELETE, `WORK_DIR` removal inherited from `setup_workdir`).
- `chmod +x`'d and passes `bash -n`.
- No embedded credentials; uses `ADMIN_USER`/`ADMIN_PASS` from the
  shared env.

## Runner-tool documentation

The four new suites use these native tools, all already known to the
release-gate runner image (see `.github/workflows/release-gate.yml`):

| Tool | Source | Status |
| --- | --- | --- |
| `gpg` (gnupg) | stock on `ak-e2e-runners` Ubuntu image | available |
| `xz` (xz-utils) | stock on `ak-e2e-runners` Ubuntu image | available |
| `ar` (binutils) | installed by `system-packages` batch (release-gate.yml:880-882) | available |
| `helm` v3+ | installed by `containers` batch (release-gate.yml:874-876) | available |
| `curl`, `jq`, `tar`, `gzip` | stock | available |

When any tool is missing the suite calls `skip_suite` with a precise
reason rather than failing the gate.

## Test plan

- [ ] `bash -n` passes on all 4 new scripts (confirmed locally)
- [ ] Suites run cleanly against a 1.1.9/1.2.0 backend with signing
  enabled on the test namespace
- [ ] Suites cleanly skip on a backend that returns 404 for
  `/api/v1/signing/keys`, `Release.gpg`, `/key`, or `Packages.xz`
- [ ] Negative controls in `test-debian-gpg-chain.sh` and
  `test-helm-provenance-verify.sh` actually catch a tampered byte
  (verified via the `EXPECT_FAILURE=1` self-test mode in `common.sh`
  if needed)
- [ ] No pollution of the runner's real GPG keyring (per-suite
  `GNUPGHOME` under `WORK_DIR`)

Refs #72, #66.

### DIFF
--- a/tests/formats/test-debian-gpg-chain.sh
+++ b/tests/formats/test-debian-gpg-chain.sh
@@ -125,6 +125,9 @@ fi
 begin_test "Create Debian repo"
 if create_local_repo "$REPO_KEY" "debian"; then
   pass
+  # Clean the repo up at suite exit so throwaway repos don't accumulate
+  # across re-runs. Mirrors the signing-key cleanup above.
+  add_exit_handler "curl -s -X DELETE -H \"\$(auth_header)\" \"\${BASE_URL}/api/v1/repositories/${REPO_KEY}\" >/dev/null 2>&1 || true"
 else
   fail "could not create debian repo"
   end_suite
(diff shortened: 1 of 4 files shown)


brandonrc:
## Code Reviewer agent: independent review

**Scope (PASS):** diff confines to `tests/formats/` four files; no `release-gate.yml` / `common.sh` edits. Clean.

**Native-tool coverage**
- PASS Debian GPG positive: `test-debian-gpg-chain.sh:313` invokes `gpg --batch --verify Release.gpg Release`.
- PASS Debian GPG negative: `test-debian-gpg-chain.sh:337-345` appends one byte and asserts non-zero exit.
- PASS Helm positive: `test-helm-provenance-verify.sh:239-244` runs `helm pull --verify --keyring`.
- WARN Helm negative weak: `test-helm-provenance-verify.sh:271-303` tampers the locally-pulled `.prov` (truncate + append `X`) and runs `helm verify` against on-disk files. Acceptable, but the comment claims the server "could hand out garbage that upload accepts and native client rejects" yet the test never re-fetches; if the backend mutates `.prov` server-side, this misses it. Consider a second `helm pull --verify` from the live URL after a re-upload attempt (or document why direct-on-disk verify is equivalent).
- PASS Tool preflight: gpg/ar/xz/helm all gate via `skip_suite` (gpg-chain:54-59, packages-xz:44-49, helm:41-46).

**OCI catalog pagination**
- PASS `?n=2` strict equality + Link rel=next: `test-oci-catalog-pagination.sh:163-175`.
- PASS strict-advance: `:204-213` uses `jq` to assert `contains_last==0` and `bad_order==0`.
- PASS past-end empty: `:234-252` asserts `length==0` and no `Link: next`.

**Packages.xz**
- PASS `xz -dc` decode + RFC822 stanza fields: `test-debian-packages-xz.sh:147-164`.
- PASS SHA256 from Release row vs on-disk: `:233-244`.
- NIT `:159` `grep -q "^SHA256: "` also matches the `SHA256:` block header line. Tighten to `^[[:space:]]+[0-9a-f]{64} ` if you want strict.

**Hygiene**
- PASS `RUN_ID` in every repo/key/alias.
- PASS Per-suite `GNUPGHOME` + `HELM_*_HOME` under `WORK_DIR` (gpg-chain:31-34, helm:49-52, 215-218).
- PASS `add_exit_handler` for `gpg-agent` kill and signing-key DELETE.
- WARN OCI suite does not register exit cleanup for the 3 created repos; re-runs accumulate (LOW: `RUN_ID` keeps keys unique).
- PASS `source common.sh`, `begin_suite/end_suite`, `pass/fail/skip` discipline correct throughout.

**FAIL Diagnostic regression**
- `test-helm-provenance-verify.sh:247-254` and `:294-302`: both `if/else` branches call `pass`, which is the same as no diagnostic. Either drop the inner `grep` (it's dead code) or make the missing-message branch a `skip`/diagnostic so we can detect helm-version regressions in stderr format.

**Verdict:** approve with two small follow-ups (dead `if/else` in helm test, optional OCI repo cleanup). No blockers.

Code Reviewer agent

brandonrc:
## Backend Architect review — PR #171

Reviewed against `artifact-keeper-api/openapi.yaml` and backend handlers (`debian.rs`, `helm.rs`, `oci_v2.rs`).

**Debian GPG chain — `test-debian-gpg-chain.sh`**
- [X] WRONG pubkey path. Backend exposes `/debian/{repo}/dists/{distribution}/gpg-key.asc` (debian.rs:61), but the test tries `/debian/{repo}/key`, `/debian/{repo}/gpg-key.asc`, then `/api/v1/signing/keys/{id}/public`. None match the format-native route; the first two 404 and the test falls back to the platform endpoint (line 266-268).
- [X] Platform endpoint returns **PEM** (X.509 `BEGIN PUBLIC KEY`), not OpenPGP (signing.rs:265, openapi.yaml:8886). `gpg --import` at line 300 will reject it, failing the chain. Add `/dists/${DISTRIBUTION}/gpg-key.asc` as the FIRST attempt.
- [OK] InRelease + Release + Release.gpg endpoint paths match (debian.rs:52-61).
- [OK] Negative tamper control is correctly structured.

**Debian Packages.xz — `test-debian-packages-xz.sh`**
- [OK] Endpoint path `/dists/{dist}/{comp}/binary-{arch}/Packages.xz` matches (debian.rs:13, line 658).
- [OK] Release manifest SHA256 row lookup is spec-correct; awk in_sha state machine is sound.
- [!] `sleep 2` for async index gen is brittle. Existing tests use the same; acceptable.

**OCI _catalog — `test-oci-catalog-pagination.sh`**
- [OK] `?n=2` strict count + Link header rel=next matches backend (oci_v2.rs:2425-2429, 301-304).
- [OK] `?last=X` strict-after semantics match SQL `WHERE (LOWER(name), name) > (LOWER($1), $1)` (oci_v2.rs:2464).
- [!] `bad_order` uses JSON `<=` (case-sensitive), backend orders by `LOWER()` then case-tiebreak. RUN_ID-suffixed lowercase keys avoid the mismatch — fragile if naming changes.
- [OK] Past-end `~zzz...` cursor sorts above all real keys; empty array + no Link is correct.

**Helm provenance — `test-helm-provenance-verify.sh`**
- [X] **No PUT route exists** in helm router (helm.rs:36-45 has only `index.yaml` GET, `charts/:filename` GET, `api/charts` POST, `api/charts/:n/:v` DELETE). Direct PUT at line 189-194 will return **405 Method Not Allowed**, not 404/501, so the test FAILS (skip_suite only catches 404/501 at line 198). Either add multipart `prov=@` fallback (as test-helm-conformance.sh:367 does) or expand the skip predicate to include 405.
- [OK] Negative control via local tamper + `helm verify` is the right code path.

**Hygiene**
- [OK] RUN_ID in all repo keys.
- [!] Debian/helm tests register signing-key DELETE on exit (gpg-chain line 110) but do NOT delete the throwaway repos. Catalog test creates 3 repos with no cleanup — these will accumulate across runs.

**Verdict:** Two blocking correctness bugs (Debian pubkey path + Helm PUT-only upload assumption). Catalog and Packages.xz tests are solid.

— Backend Architect agent

brandonrc:
Addressed Backend Architect findings (commit 71d2094):

- **`test-debian-gpg-chain.sh`**: Fetch the OpenPGP public key from the correct route `/debian/{repo}/dists/{distribution}/gpg-key.asc` (per `debian.rs:61`). Removed the X.509 `/api/v1/signing/keys/{id}/public` fallback — it returns a `BEGIN PUBLIC KEY` PEM that `gpg --import` rejects, breaking the entire apt-secure trust chain. Now skip-with-reason if the OpenPGP endpoint truly isn't available, and fail loudly if it returns X.509 instead of OpenPGP.
- **`test-helm-provenance-verify.sh`**: Switched to a single multipart `POST /helm/{repo}/api/charts` with `chart=@<tgz>` and `prov=@<prov>` fields (pattern from `test-helm-conformance.sh:367`). The previous direct `PUT /charts/*.tgz.prov` returned 405 because the Helm router (`helm.rs:35-45`) has no PUT route for `.prov`, and the 404/501 skip guard didn't catch it.
- **Hygiene cleanup**: Added a `cleanup_repo` exit handler (`curl -X DELETE -H "$(auth_header)" "${BASE_URL}/api/v1/repositories/${REPO_KEY}"`) immediately after each `create_local_repo` call in all three new suites (`test-debian-gpg-chain.sh`, `test-helm-provenance-verify.sh`, `test-oci-catalog-pagination.sh`) so throwaway repos no longer leak across re-runs. Mirrors the existing signing-key cleanup pattern.

All three modified scripts pass `bash -n`.

Issue #69, opened 2026-04-27
Epic 1: Pull-through cache reliability E2E coverage (v1.1.9)


brandonrc:
Direct response to customer @Firjens (#872): "the two features that make artifact-keeper compelling -- the pull-through cache and security scanning -- are unfortunately the ones with the most issues."

**Policy:** every gap below is a 1.1.x feature; tests must land before v1.1.10. Tracked under coverage policy #66.

## Sub-tasks

1.1 Concurrent fetch stampede prevention (proxy_max_concurrent_fetches semaphore, queue_timeout)
1.2 Cache poisoning full E2E (current test-cache-poisoning.sh skips the load-bearing assertion)
1.3 Stale cache fallback when upstream is down (get_stale_cached_artifact)
1.4 Cache TTL expiry behavior, per-repo TTL, refresh
1.5 ETag-based cache validation (304 responses, conditional requests)
1.6 OCI bearer token caching with TTL eviction
1.7 OCI 401-redirect token-realm SSRF validation
1.8 Max artifact size streaming enforcement
1.9 Format-specific upstream auth (private registry mirroring with creds)
1.10 Virtual repo proxying across multiple upstreams (cross-format generic test)

## Out of scope (1.2.0 dependency)

None -- all of the above are existing 1.1.x features.

brandonrc:
## Status of this epic after PR #152

PR #152 (cluster D) lands the first wave of pull-through cache reliability coverage. Three of the ten sub-tasks now have shipped tests; the remaining seven need follow-up sub-issues.

### Shipped in #152

- 1.2 Cache poisoning class -- partial. `tests/pullthrough/test-cache-hit-no-refetch.sh` pins the count-axis assertion (N consecutive fetches return byte-identical bodies in the TTL window). Full poisoning E2E is still blocked on backend artifact-keeper#1224 (AK_SSRF_ALLOW_PRIVATE_CIDRS env var) per the existing `tests/security/test-cache-poisoning.sh` skip header.
- 1.4 Cache TTL behavior -- two complementary scripts now in place: the existing `tests/repos/test-cache-ttl-eviction.sh` (time-axis: TTL expiry refetches upstream) plus `tests/pullthrough/test-ttl-expiry-refetch.sh` (two-cycle: state machine doesn't get stuck after first eviction) plus `tests/pullthrough/test-cache-hit-no-refetch.sh` (within-TTL stability).
[...]

PR #173, opened 2026-05-16
test(cache): stampede + stale fallback + ETag + OCI SSRF (#69)

brandonrc:
## Summary

Adds 4 pull-through cache E2E tests under `tests/pullthrough/` targeting the load-bearing assertion classes called out in issue #69 (Epic 1, customer pain #872):

- **1.1 Concurrent fetch stampede prevention** -- N=10 concurrent client requests on a cold cache. Asserts byte-identical 2xx bodies (coalesce) + 503-or-2xx-only status set. Feature-gated to `proxy_stampede_protection` (>=1.2.0).
- **1.3 Stale-on-upstream-error** -- prime cache, DELETE upstream repo, assert R still serves the cached body (not 5xx, not 404). Exercises `proxy_service.rs:get_stale_cached_artifact`.
- **1.5 ETag conditional request** -- conda `repodata.json` returns ETag; matching `If-None-Match` -> 304 empty body; wrong / wildcard / multi-value variants all behave per RFC 7232.
- **1.7 OCI remote upstream SSRF** -- POST `/api/v1/repositories` with `format=docker` and `upstream_url` pointing at IMDS / loopback / RFC1918 / cluster DNS / IPv6 link-local must be rejected. Includes a legitimate-external control case so the assertion is not vacuous against a reject-everything regression.

## Why these four

The issue lists 1.1-1.10. Selected based on (a) load-bearing assertions the customer ticket names directly and (b) testability without the mock-upstream fixture, which is currently SSRF-blocked pending `artifact-keeper#1224` (see `tests/security/test-cache-poisoning.sh` for the chicken-and-egg).

All four use the AK-to-AK pattern (two local repos, backend dials itself) already proven in `test-cache-ttl-eviction.sh` and `test-cache-hit-no-refetch.sh`. No mock-upstream dependency.

## Deferred (explicitly out of scope for this PR)

- **1.2** cache poisoning -- existing `tests/security/test-cache-poisoning.sh` stub holds the surface; full E2E blocked on `artifact-keeper#1224`.
- **1.4** cache TTL -- covered by `tests/repos/test-cache-ttl-eviction.sh` + `tests/pullthrough/test-ttl-expiry-refetch.sh`. No new test here would add assertion power.
- **1.6** OCI upstream-token caching -- requires observing `/v2/token` request counts to an upstream, which needs mock-upstream.
- **1.8** max artifact size -- the limit is a global system setting (`max_upload_size_bytes`); temporarily lowering it conflicts with parallel suites.
- **1.9 / 1.10** -- format-specific upstream auth and virtual-repo cross-format proxying; larger surfaces, separate PRs.

## Test plan

- [ ] `bash -n` clean on all 4 scripts (verified locally)
- [ ] All 4 are chmod +x (verified locally)
- [ ] Run via `./scripts/run-suite.sh --suite pullthrough --run-id local-$(date +%s)` against a local backend
- [ ] Confirm release-gate `pullthrough-tests` job picks up the new scripts (auto-discovered by `find tests/pullthrough -name 'test-*.sh'`)
- [ ] Verify the stampede test gracefully skips on a 1.1.x backend via `require_feature proxy_stampede_protection`
- [ ] Verify the SSRF test's control case is accepted on a backend that allows external upstream URLs
- [ ] Verify EXIT handlers clean up created repos (no orphan repos after a failed run)

### DIFF
--- a/tests/pullthrough/test-cache-stampede-no-upstream-divergence.sh
+++ b/tests/pullthrough/test-cache-stampede-no-upstream-divergence.sh
@@ -1,17 +1,44 @@
 #!/usr/bin/env bash
-# test-cache-stampede-prevention.sh -- Concurrent-fetch stampede must
-# NOT produce N upstream fetches for N client requests on a cold cache
+# test-cache-stampede-no-upstream-divergence.sh -- Concurrent fan-out
+# under a stampede MUST NOT produce divergent results served to clients
 # (issue #69 sub-task 1.1).
+#
+# Naming note (and what this test actually proves)
+# ------------------------------------------------
+# This file was originally named test-cache-stampede-prevention.sh.
+# That name overclaimed: the load-bearing assertions below (byte-
+# identical 2xx response bodies + 503-or-2xx-only status set) are
+# NECESSARY for a correctly-coalesced proxy, but they are NOT SUFFICIENT
+# to prove coalescing on their own. A non-coalesced backend that happens
+# to hit a deterministic local-PyPI upstream N times will also pass this
+# test, because each of the N upstream fetches returns the same bytes.
+# The real per-key semaphore admit count can only be proved by a
+# source-side fetch counter, which lives on the mock-upstream peak-
+# inflight signal and is currently SSRF-blocked pending artifact-keeper
+# #1224 (see the dedicated stub in tests/security/test-cache-stampede.sh).
+#
+# So the name was corrected to reflect what is actually proved here:
+# "no upstream divergence under stampede" -- i.e. regardless of how the
+# N concurrent fetches are scheduled, every successful client sees the
+# same bytes and any admission-control rejection surfaces as a clean
+# 503 rather than a 5xx crash. When #1224 lands and the mock unstubs,
+# the count-based sibling in tests/security/ takes ownership of the
+# "prevention" assertion; this script keeps the divergence-equality
+# assertion as a complementary check.
 #
 # What this asserts
 # -----------------
 # proxy_service in the backend exposes proxy_max_concurrent_fetches
-# (default 20) plus proxy_queue_timeout_secs (default 30). The contract:
-# N concurrent client requests for the SAME (repo, path) tuple on a
-# cold cache MUST be coalesced -- only one (or at most M, where M is
-# the per-key semaphore admit) upstream fetch happens, and the other
-# N-1 / N-M callers either wait for the in-flight fetch and reuse its
-# result or time out with 503.
+# (default 20) plus proxy_queue_timeout_secs (default 30). The contract
+# this test exercises is the CLIENT-OBSERVABLE half of stampede handling:
+# regardless of how many of the N concurrent client requests for the
+# SAME (repo, path) tuple on a cold cache actually reach the upstream,
+# every 2xx response served to a client MUST be byte-identical, and any
+# admission-control rejection MUST surface as 503 (not a 5xx crash and
+# not a silently different body). Coalescing collapses N fetches to 1
+# and trivially passes this contract; a non-coalesced run can also pass
+# it if the upstream is deterministic, but a coalesced run can never
+# FAIL it.
 #
 # Why this matters
 # ----------------
@@ -84,7 +111,7 @@
 
 source "$(dirname "$0")/../lib/common.sh"
 
-begin_suite "cache-stampede-prevention"
+begin_suite "cache-stampede-no-upstream-divergence"
 auth_admin
 setup_workdir
 
--- a/tests/pullthrough/test-cache-stampede-prevention.sh
+++ b/tests/pullthrough/test-cache-stampede-prevention.sh
@@ -1,0 +1,338 @@
+#!/usr/bin/env bash
+# test-cache-stampede-prevention.sh -- Concurrent-fetch stampede must
+# NOT produce N upstream fetches for N client requests on a cold cache
+# (issue #69 sub-task 1.1).
+#
+# What this asserts
+# -----------------
+# proxy_service in the backend exposes proxy_max_concurrent_fetches
+# (default 20) plus proxy_queue_timeout_secs (default 30). The contract:
+# N concurrent client requests for the SAME (repo, path) tuple on a
+# cold cache MUST be coalesced -- only one (or at most M, where M is
+# the per-key semaphore admit) upstream fetch happens, and the other
+# N-1 / N-M callers either wait for the in-flight fetch and reuse its
+# result or time out with 503.
+#
+# Why this matters
+# ----------------
+# Without coalescing, every CI runner that wakes up at midnight to
+# pull the same nightly tarball produces N parallel upstream fetches
+# (we have seen 600+ on a single popular blob). For an upstream rate-
+# limited at 100 req/min this guarantees a thundering-herd failure on
+# the first cold-cache fetch every day. This is the second-class of
+# customer pain in artifact-keeper#872.
+#
+# Existing coverage gap
+# ---------------------
+# tests/security/test-cache-stampede.sh exists but is hard-stubbed
+# pending backend issue artifact-keeper#1224 (AK_SSRF_ALLOW_PRIVATE_CIDRS
+# so the mock fixture can be reached from the runner pod). That test
+# uses the mock-upstream peak-inflight counter as its primary signal.
+# This new test takes a different approach that does NOT require the
+# mock: AK-to-AK (same pattern as test-cache-hit-no-refetch.sh +
+# test-cache-ttl-eviction.sh) plus a content-equality observable.
+#
+# Approach
+# --------
+# Two-repo AK-to-AK:
+#   U = local PyPI repo we publish v1 to.
+#   R = remote PyPI repo whose upstream_url is U's AK URL.
+#
+# Then:
+#   1. Fan out N concurrent fetches through R's /simple/<pkg>/ on a
+#      cold cache (R was just created -- nothing cached).
+#   2. Collect (a) all HTTP statuses and (b) all response bodies.
+#   3. Load-bearing assertions:
+#      (i)  All non-error responses MUST have byte-identical bodies.
+#           If the stampede produced N independent upstream fetches
+#           and the upstream's simple-index generator races (it
+#           often does -- the html shape can change between requests
+#           under load), we'd see body drift.
+#      (ii) Either:
+#           - ALL N responses succeeded (2xx), OR
+#           - Some responses are 503 Service Unavailable (queue
+#             timeout exceeded -- this is the proxy_queue_timeout_secs
+#             admission control firing correctly). Either outcome
+#             is a PASS; what fails is mixed 2xx-with-different-bodies
+#             (stampede actually happened) or any 5xx other than 503
+#             (a fetch crashed mid-coalesce).
+#
+# What this does NOT assert (deferred to stubbed mock-upstream test)
+# ------------------------------------------------------------------
+# - Peak upstream-fetch count <= proxy_max_concurrent_fetches.
+#   That requires the mock-upstream peak-inflight counter, which is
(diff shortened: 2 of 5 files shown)


brandonrc:
## Code Review: PR #173 (test-author lens)

**Scope:** Clean — `gh pr diff --name-only` shows only the four `tests/pullthrough/*.sh` files. No `release-gate.yml`/`common.sh` drift. ✅
**CI wiring:** `release-gate.yml:1029-1063` runs `run-suite.sh --suite pullthrough`, which globs `tests/pullthrough/**/test-*.sh` — new files auto-discovered. ✅
**Helpers:** All four source `tests/lib/common.sh`, use `RUN_ID` in resource keys, register `add_exit_handler` cleanup, and use `auth_header`/`format_auth_header` correctly for admin-vs-format routes. ✅

### Per-file

**test-cache-stampede-prevention.sh**
- ✅ Feature-gates on `proxy_stampede_protection` (`feature-flags.sh:95`) before any setup — skip-vs-fail discipline correct.
- ✅ AK-to-AK pattern avoids the SSRF-blocked mock (per the repo's documented constraint).
- ⚠ Load-bearing assertion is weaker than the header claims: byte-equality across N 2xx responses is NECESSARY but not SUFFICIENT — a non-coalesced backend hitting a deterministic local-PyPI upstream N times will also pass with identical bodies. The comment at L60-67 acknowledges this and defers the peak-count assertion to the (currently stubbed) mock-upstream sibling, which is the right call given constraints, but the test's name oversells what it catches. Consider renaming the test description to "no body-drift under concurrent fetch" to match what's actually proved.
- ✅ Assertion #3 (post-burst cache-write check at L309-326) is the strongest part.

**test-stale-on-upstream-error.sh**
- ❌ **No `require_feature` gate.** No flag exists for stale-on-error in `feature-flags.sh`. On a backend without `proxy_service.rs` stale-fallback wiring the test will hard-FAIL at L259 — i.e., it cannot distinguish "feature missing" from "feature regressed." Add a `stale_on_upstream_error` token to `AK_BACKEND_BRANCH_1_2_X` (or wherever the wiring actually landed) and gate at L66, mirroring the stampede pattern.
- ✅ Pre-delete guard at L194-198 (`grep -q "${PKG_VERSION}"`) defends against the empty-prime confounder.
- ✅ 404-sanity-after-delete at L239-247 distinguishes "stale fired" from "U still up."
- ✅ Error classification at L264-274 is precise — separates 200-empty, 404-passthrough, and 5xx-propagation.

**test-etag-conditional-request.sh**
- ❌ **No `require_feature` gate.** Same problem as stale-on-error: header at L7-14 names `conda.rs:391-432` but no feature flag enforces backend version. Older backend → hard fail instead of skip. Add a flag (e.g. `conda_etag_conditional`) and gate.
- ✅ Both legs covered (200+ETag at L134-157; 304-on-match at L168-187; 200-on-mismatch at L195-213 prevents constant-true vacuous pass).
- ✅ ETag syntax regex at L150 (`^"[0-9a-f]{64}"$`) enforces RFC 7232 quoted form.
- 💭 L181: `wc -c` on 304-body verifies "empty" but conda.rs may emit a single trailing newline depending on framing; consider `[ "$(wc -c < "$body_file")" -le 1 ]` for tolerance. Low priority.

**test-oci-remote-upstream-ssrf.sh**
- ✅ Six attack vectors (IMDS, loopback, RFC1918 x2, cluster DNS, IPv6 link-local) — strong coverage.
- ✅ Control case at L186-202 is the right defense against false-pass-from-reject-all.
- ⚠ Control URL `https://registry-1.docker.io/v2/` (L187) is fine at create-time (no probe), but if backend ever adds a connectivity check at create, this will start pulling Docker Hub on every release-gate run. The L181-183 comment already calls this out — acceptable, but consider parameterizing via env var so a workflow can swap to an internal mirror.
- ✅ `try_create_remote_oci` deliberately bypasses `create_remote_repo`'s retry (L86-88) to capture the RAW reject — correct, retries would mask 400/422.
- ✅ Critical-finding cleanup at L135 (delete-on-accept) prevents leftover SSRF-primitive repos.

### Cross-cutting

- ⚠ All four files duplicate the `EXPECT_FAILURE` post-amble (e.g. stampede L328-336) even though `common.sh:976` `end_suite` already handles it. Calling `end_suite` in a subshell + again unconditionally writes JUnit XML twice (second overwrites first). **However**, this matches existing convention in `tests/pullthrough/test-cache-hit-no-refetch.sh:272+` and `test-virtual-shadowing-guard.sh:463+`, so it's a repo-wide style issue, not a PR-173 regression. Out of scope to fix here; worth a follow-up cleanup task.
- ✅ All four exit fast on `skip_suite` for upload-endpoint unavailability — no vacuous passes.
- ✅ No public-internet reach except the documented OCI control case.

### Verdict

**Approve with two blockers:** add `require_feature` gates to `test-stale-on-upstream-error.sh` and `test-etag-conditional-request.sh` (or document why those features are universal across the supported backend matrix). The stampede assertion-strength caveat and the EXPECT_FAILURE post-amble are non-blocking.

— Code Reviewer agent

brandonrc:
## Backend Architect review (correctness, not style)

**1.1 Stampede (`test-cache-stampede-prevention.sh`)**
- ❌ Premise mismatch: `proxy_service.rs:98` constructs a **global** `Semaphore(proxy_max_concurrent_fetches=20)` — NOT a per-(repo,path) coalescer. `fetch_from_upstream` acquires at `proxy_service.rs:403` and the test's claim "N concurrent same-key requests MUST be coalesced" is unsupported by the code. With N=10 < 20 default, every request will proceed in parallel and each does its own upstream fetch.
- ⚠ Body-equality assertion is near-vacuous: the upstream is a freshly-published single-package PyPI simple-index; the HTML is deterministic across repeated reads. Test would PASS even if no coalescing occurred. The peak-inflight assertion (deferred to mock-upstream) is the only one that actually detects the bug.
- ⚠ `proxy_stampede_protection` flag (1.2.0 in `tests/lib/common.sh:217`) gates correctly, but the underlying feature it gates on doesn't appear to exist as described in `proxy_service.rs` HEAD — verify the flag isn't documenting a not-yet-shipped capability.

**1.3 Stale-on-error (`test-stale-on-upstream-error.sh`)**
- ✅ Path is correct: `proxy_service.rs:185-198` Err arm → `get_stale_cached_artifact` (`:853`). No feature flag — always on. AK-DELETE → 404 from upstream → `fetch_from_upstream` returns `Err(Storage(...))` → stale fallback fires.
- ⚠ `X-Cache: STALE` is in `build_stale_cache_headers` (`:1032-1037`) but the `fetch_artifact_with_cache_path` Err arm at `:194` returns `(content, content_type)` only — headers are not propagated to the client. Bonus assertion would be vacuous; test correctly doesn't gate on it.

**1.5 ETag (`test-etag-conditional-request.sh`)**
- ✅ `conda.rs:391 compute_etag` emits `"<sha256>"` (strong validator, RFC 7232 §2.3). `:400 check_conditional_request` handles `*`, comma-split, trim. Spec-correct.
- ⚠ Conda-only — handler in `conda.rs:435` `build_cacheable_response`. Other formats (`pypi.rs`, `maven.rs`) do NOT emit ETag. Header docstring is honest about this.

**1.7 OCI SSRF (`test-oci-remote-upstream-ssrf.sh`)**
- ❌ Case 5 (`kubernetes.default`) will FAIL on most clusters: `validation.rs:57 BLOCKED_HOSTS` does NOT include `kubernetes.default`, no DNS resolve, so `validate_outbound_url` accepts the hostname at create time. The test will see 2xx → fail. Either add `kubernetes.default` to BLOCKED_HOSTS or drop that case.
- ✅ IPv4 cases (169.254, 127.0.0.1, 10.x, 192.168.x) and IPv6 fe80:: are all covered by `is_blocked_ipv4`/`is_blocked_ipv6` (`:181-226`). `AppError::Validation` → 400 (`error.rs:94`).
- ⚠ Control case `registry-1.docker.io`: cluster egress to public internet is NOT inherently blocked (only RFC1918 mock-upstream is, per `test-cache-poisoning.sh:9-18`). Control should succeed when egress is allowed; otherwise `skip` keeps the suite honest. Acceptable.

**Cross-cutting**
- ✅ RUN_ID hygiene clean; per-suite cleanup via `add_exit_handler`.
- ✅ Auto-discovered by `.github/workflows/release-gate.yml:1063`.
- ⚠ All four `EXPECT_FAILURE=1` blocks call `( end_suite )` in a subshell — `end_suite` writes JUnit to disk, which subshell-isolation may silently lose. Verify locally.

**Verdict:** 1.3 and 1.5 are correct. 1.1 tests a coalescing property the backend does not implement (will pass for the wrong reason). 1.7 case 5 will fail against current `BLOCKED_HOSTS`. Both warrant action before merge.

— Backend Architect agent

brandonrc:
Addressed Code Reviewer findings (commit 5332427):

- **BLOCKER `test-stale-on-upstream-error.sh`**: added a feature gate at suite start. Probes `get_backend_version` and `skip_suite`s when backend < 1.2.0 (the proxy correctness epoch alongside `proxy_stampede_protection` / `proxy_ttl_eviction_correctness`), and also `skip_suite`s when `/api/v1/repositories` returns 404/501. Used the in-suite `skip_suite` pattern (see `test-cache-hit-no-refetch.sh`, `test-multipart-artifact-upload.sh`) rather than introducing a new `require_feature` token, since that would require touching `tests/lib/common.sh` which a separate change owns.

- **BLOCKER `test-etag-conditional-request.sh`**: same approach. Probes `/conda/<bogus>/<bogus>/repodata.json` at suite start and `skip_suite`s on 405/501 (handler absent) or connect-timeout. 401/403/404/200/304 all confirm the conda handler is mounted and let the suite proceed.

- **SOFT `test-cache-stampede-prevention.sh` -> `test-cache-stampede-no-upstream-divergence.sh`**: picked option (a) and renamed (file + suite name + header). The source-side fetch count is infeasible from this script (SSRF-blocked mock-upstream pending artifact-keeper#1224); ownership stays with `tests/security/test-cache-stampede.sh`. Header now spells out that the byte-equality assertion is NECESSARY but NOT SUFFICIENT for coalescing and explains the corrected scope.

All three files pass `bash -n`. `tests/lib/common.sh` and `.github/workflows/release-gate.yml` were not modified.

brandonrc:
Addressed Backend Architect findings on this PR:

- **Finding #2 (OCI SSRF, case 5)**: Replaced `kubernetes.default` with `metadata.google.internal` in `tests/pullthrough/test-oci-remote-upstream-ssrf.sh`. The original hostname is NOT in `BLOCKED_HOSTS` (`backend/src/api/validation.rs:57-69`: `localhost`, `metadata.google.internal`, `metadata.{azure,tencentyun,oraclecloud,platformequinix}.com`, `backend`, `postgres`, `redis`, `meilisearch`, `trivy`), so a correctly-defended backend would 2xx-accept it and fail the assertion. `metadata.google.internal` IS in the list and preserves the case's original intent (exercise the DNS-name path of the blocklist rather than the IP-literal path). Kube-internal service SSRF is still covered by the RFC1918 / loopback IP cases and by the docker-internal hostnames (`backend`, `postgres`, etc.) that ARE in `BLOCKED_HOSTS`.

- **Finding #1 (cache stampede test, global vs per-key semaphore)**: No additional change needed. The in-flight code-review fixup (5332427) already applied option (b) -- renamed `test-cache-stampede-prevention.sh` to `test-cache-stampede-no-upstream-divergence.sh` and rewrote the header + load-bearing claims to make explicit that `fetch_semaphore` in `proxy_service.rs:98` is a global `Semaphore(max_concurrent)`, NOT a per-`(repo, path)` coalescing primitive. The body-equality assertion is now correctly framed as necessary-but-not-sufficient for coalescing, with the count-based "prevention" assertion deferred to the stubbed `tests/security/test-cache-stampede.sh` (pending `artifact-keeper#1224`).

Both files pass `bash -n`. Pushed as 8dfdb62.

Issue #179, opened 2026-05-19
Verify #152 pullthrough/security suites against a real backend before tagging v1.1.9


brandonrc:
## Background

[#152](https://github.com/artifact-keeper/artifact-keeper-test/pull/152) merged 2026-05-16 with 2,882 lines of new test code across `tests/pullthrough/`, `tests/security/`, `tests/platform/`, plus new release-gate workflow jobs (`pullthrough-tests`, `security-tests`). 

**These new suites were never executed against a backend in CI before merge.** The PR's only pre-merge checks were:
- `bash -n` syntax check
- `shellcheck`
- GitGuardian
- 2 self-test fixture jobs from earlier waves (smoke + scan-completion)

The new pullthrough-tests and security-tests jobs run on `workflow_dispatch` with `inputs.backend_tag` — which was never triggered for this PR. The PR body's `Test plan` checkbox

> Run the new suite end-to-end against the latest release/1.1.x deploy

is unchecked.

## What's needed

A one-off `workflow_dispatch` run of `release-gate.yml` against a recent backend tag (`1.1.8`, `1.1.9-rc.N`, or the current `release/1.1.x` build) to confirm:

1. The new pullthrough-tests job actually runs (deploy succeeds, suite executes, doesn't time out)
2. The new security-tests job actually runs
3. `test-feature-flag-drift.sh` matches the deployed backend version (no version skew)
4. All new tests pass against a real backend, OR failures are captured as follow-up issues

## Why this matters now

Tagging v1.1.9 with untested test infrastructure is the same silent-success class the gate exists to catch (#870/#871/#888) — except now the silent failure is in the gate itself.

## Suggested approach

```
gh workflow run release-gate.yml --repo artifact-keeper/artifact-keeper-test \
    -f backend_tag=1.1.9-rc.N \
    -f test_suite=all \
    --ref main
```

Capture the run URL and the per-job outcomes in this issue's comments before closing.

Note: #176 (Tier 2b follow-up review) fixes 4 quality bugs in #152's merged code (TEST_TIMEOUT, /tmp hygiene, glob anchoring, observation count). Run this verification AFTER #176 lands so the fixes are included.

## Refs

- #152 (the PR that needs end-to-end validation)
- #176 (Tier 2b fixes, prerequisite)
- artifact-keeper#870, #871, #888 (silent-success class this gate guards against)

brandonrc:
[...]

PR #183, opened 2026-05-19
test(security): SQL-injection LIKE-wildcard guard via helm (closes #181)

### LINKED ISSUE #181: test: SQL-injection LIKE-wildcard regression guard (#1004 follow-up)
## Background

PR [#180](https://github.com/artifact-keeper/artifact-keeper-test/pull/180) attempted a regression guard for artifact-keeper#1004 (SQL injection via unescaped LIKE wildcards in 15 call sites). The 2-judge review pass caught the test as a fantasy-pass: it exercised \`/api/v1/repositories/{key}/artifacts/{path}\`, which is the REPOSITORIES CRUD handler. That handler uses exact-equality SQL:

\`\`\`sql
WHERE repository_id = \$1 AND path = \$2
\`\`\`

It does NOT call \`local_fetch_by_path_suffix\` or any of the 15 LIKE call sites fixed in artifact-keeper#1004. The wildcard probes returned 404 pre-fix and post-fix identically — no signal about the actual LIKE-injection surface.

## What's needed

A test that hits a native-format handler taking a user-supplied filename through one of the named LIKE sites. Cleanest candidates from the #1004 fix list:

| Format | Function | URL pattern |
|--------|----------|-------------|
| **pypi** | \`serve_file\` (escape_filename_for_like) | \`/pypi/{repo}/simple/{name}/{filename}\` |
| **helm** | \`download_chart\` (escape_filename_for_like, absorbed from #984) | \`/helm/{repo}/charts/{chart}-{version}.tgz\` |
| **rubygems** | \`download_gem\` (escape_like_literal, absorbed from #984) | \`/rubygems/{repo}/gems/{filename}\` |
| **npm** | \`serve_tarball\` (escape_like_literal x2, line 848) | \`/npm/{repo}/{name}/-/{filename}\` |

Recommended starting point: **helm** because the upload protocol is the simplest of the four (single PUT of a .tgz with no signing/multipart).

## Test sketch

\`\`\`bash
# 1. Create local helm repo
# 2. Upload chart 'pkg-target-RUNID-1.0.0.tgz'
# 3. Upload decoy 'pkg-decoy-RUNID-1.0.0.tgz'
# 4. Sanity: GET /helm/{repo}/charts/pkg-target-RUNID-1.0.0.tgz returns the planted bytes
# 5. Wildcard probes against /helm/{repo}/charts/...:
#    a. pkg-_arget-RUNID-1.0.0.tgz (underscore matches literal 't')
#    b. pkg-%25.tgz (percent matches the rest)
#    c. %25kg-target-RUNID-1.0.0.tgz (percent matches 'p')
# 6. Each probe MUST NOT return the planted bytes (404 or different content)
# 7. Cleanup
\`\`\`

## Negative control

Per the judge's suggestion, add a probe whose wildcards target a name that genuinely does not exist; pre-fix AND post-fix should both return 404. The positive test then has signal only when paired with a probe that would have matched pre-fix.

## Refs

- artifact-keeper#1004 \`18b1fa87\` (corrected SHA; the squash-merge commit on origin/main)
- artifact-keeper#874 \`d7a60307\` (the calibration's surviving test, landed in #180)
- [#180](https://github.com/artifact-keeper/artifact-keeper-test/pull/180) review thread (judge findings)

brandonrc:
## Summary

Closes [#181](https://github.com/artifact-keeper/artifact-keeper-test/issues/181), which itself was filed from PR [#180](https://github.com/artifact-keeper/artifact-keeper-test/pull/180)'s judge review after the original generic-format SQL-injection test was identified as a fantasy-pass: the generic repositories CRUD route uses exact-equality SQL (`WHERE path = $2`) and never reaches a LIKE clause.

Guards artifact-keeper#1004 (`18b1fa87`, merged 2026-05-04), which extended the LIKE-wildcard escape to 15 call sites. Pre-fix, the helm chart download path `helm.rs::download_chart` interpolated the URL filename segment into a SQL LIKE pattern without escaping `%` or `_`. An attacker who could read one chart under a repo could read any other chart by crafting URLs with those wildcards.

## Why helm

Of the 15 fixed sites, helm has:
- Simplest upload protocol (ChartMuseum-compatible POST multipart on `/helm/{repo}/api/charts`) — no signing, no multipart assembly noise
- Download URL `/helm/{repo}/charts/{filename}.tgz` puts the entire filename segment through `download_chart`'s LIKE clause — exactly the bytes the bug allowed to wildcard-match

## Test shape

1. Create local helm repo
2. Package + upload TWO charts: planted (`pkg-target-RUNID-1.0.0`) and decoy (`pkg-decoy-RUNID-1.0.0`). The decoy is what wildcards might *legitimately* match — its presence pins that the bug-class under test is "PLANTED bytes leak via wildcard" specifically.
3. **Sanity**: exact-match GET returns planted chart bytes (SHA-256 equality). If this fails, `skip_suite` — wildcard assertions would be meaningless.
4. **Negative control**: a never-uploaded filename returns 4xx. If 200, the environment has a catch-all matcher and the test isn't exercising #1004.
5. **Four wildcard probes**, each crafted so pre-fix the LIKE pattern matches the planted chart:
   - `pkg-_arget-RUNID-1.0.0.tgz` (underscore mid-name)
   - `pkg-target-RUNID-%25.tgz` (percent on version suffix)
   - `_kg-target-RUNID-1.0.0.tgz` (underscore at start)
   - `%25kg-target-RUNID-1.0.0.tgz` (percent at start)

Each MUST NOT return the planted-chart SHA-256. 404, 400, or 200 with non-planted bytes (e.g. the decoy) are all post-fix acceptable outcomes.

## Verification

- [x] `bash -n` clean
- [ ] Local docker-compose exec deferred (backend in docker-desktop-extension network with internal-only port mapping; release-gate workflow_dispatch is the authoritative execution path, tracked separately via [#179](https://github.com/artifact-keeper/artifact-keeper-test/issues/179))
- [ ] Smoke self-test fixtures auto-run on this PR

## Refs

- [#180](https://github.com/artifact-keeper/artifact-keeper-test/pull/180) — calibration that surfaced the fantasy-pass problem
- [#181](https://github.com/artifact-keeper/artifact-keeper-test/issues/181) — the issue this PR closes
- artifact-keeper#1004 `18b1fa87` — the fix being guarded
- artifact-keeper#984 — the never-merged predecessor #1004 absorbed; helm was on its fix list

### DIFF
--- a/tests/security/test-sql-injection-like-wildcards-helm.sh
+++ b/tests/security/test-sql-injection-like-wildcards-helm.sh
@@ -83,6 +83,12 @@ WILDCARD_PROBES=(
 
 begin_test "Create helm local repo"
 if create_local_repo "$REPO_KEY" "helm"; then
+  # Register server-side cleanup via add_exit_handler so the repo is
+  # removed even if the script aborts mid-way (set -e tripping on
+  # `helm package` failure, an unexpected `curl` exit, etc.). Linear
+  # cleanup at end-of-file would leak the repo on the backend in
+  # those cases.
+  add_exit_handler "api_delete /api/v1/repositories/${REPO_KEY} >/dev/null 2>&1 || true"
   pass
 else
   fail "could not create helm repo (${REPO_KEY})"
@@ -90,6 +96,10 @@ fi
 
 # Build two minimal but distinct charts so wildcard matching has
 # something to potentially-incorrectly resolve to.
+#
+# Returns 0 on success, 1 on failure. We avoid letting `set -e` rip
+# the script down on `helm package` failure because the test framework
+# wants a structured `fail` row in the JUnit output, not a bare exit.
 build_chart() {
   local chart_name="$1"
   local version="$2"
@@ -116,7 +126,10 @@ metadata:
 data:
   sigil: {{ .Chart.Description | quote }}
 EOF
-  helm package "${chart_dir}" -d "${WORK_DIR}" >/dev/null
+  if ! helm package "${chart_dir}" -d "${WORK_DIR}" >/dev/null 2>&1; then
+    return 1
+  fi
+  return 0
 }
 
 upload_chart() {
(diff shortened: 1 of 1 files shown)


brandonrc:
Two-judge pass complete (per the goal file's worker+judge pattern).

**Reality Checker — APPROVE**
Traced the actual LIKE call site: `helm.rs::download_chart` (L389-401 on origin/main) → `proxy_helpers::find_local_by_filename_suffix` (L1501-1517) which runs `SELECT ... WHERE path LIKE '%/' || $2 ESCAPE '\'` with `escape_like_literal($2)`. This IS one of the call sites #1004 hardened, so the test exercises the real fixed surface (not a fantasy-pass like the original generic-route attempt in #180). All four wildcard probes verified as pre-fix matches:
- `pkg-_arget-...` — `_` matches literal `t`
- `pkg-target-...-%25.tgz` — `%` matches `1.0.0`
- `_kg-target-...` — `_` matches `p`
- `%25kg-target-...` — `%` matches `pkg`

Negative control (`never-was-here-RUNID-9.9.9.tgz` → must 4xx, else `skip_suite`) is sound and protects against catch-all-route environments.

**Code Reviewer — LGTM-WITH-NITS**
Three nits applied in `526b5ad`:

1. `helm package` wrapped in `if ! helm package; then return 1; fi` so a packaging failure surfaces as a structured `fail()` row instead of silent set-e exit
2. Probe scratch files moved from `mktemp -t` (TMPDIR) to `mktemp -p "$WORK_DIR"` so the `setup_workdir` EXIT handler always reaps them
3. Server-side repo cleanup registered via `add_exit_handler` immediately after `create_local_repo` succeeds, so failure between create and end-of-file no longer orphans the repo on the backend

**Non-blocking suggestions from judges** (not applied):
- A 5th probe combining `%` and `_` for completeness — kept the 4-probe set, all bug-class coverage is present
- PR body could clarify that #1004 the *merged PR* contains the helm fix (commit `18ef2531` absorbed it from never-merged #984); strictly the original #1004 commit `9ad27f06` was 10 sites — the merged-PR shape this guards is the 15-site shape
- Run end-to-end with the fix reverted to lock in the pre-fix RED state once #179 lands the execution mechanism

PR #184, opened 2026-05-19
test(security): SSRF allowlist guard for Go sumdb proxy (#879)

brandonrc:
## Summary

Guards artifact-keeper#879 (`3e529938`, merged 2026-04-25): the Go proxy's sumdb route used to build `https://{host}/{path}` from URL path components controlled entirely by the caller. Textbook SSRF — original repro was

```
GET /go/{repo}/sumdb/169.254.169.254/latest/meta-data/iam/security-credentials/
```

which made the server fetch AWS/GCP IMDS on the attacker's behalf and return the body.

Fix: static allowlist of `sum.golang.org` and `sum.golang.google.cn`. Any other host returns 403. Comparison case-insensitive per RFC 1035.

Continues the worker+judge pattern from #180/#183 — pure URL probes against a local-format repo (no upload needed), so the regression-of-interest is isolated from format-handler noise.

## Coverage (10 cases)

| Surface | Hosts | Expected |
|---|---|---|
| Cloud metadata | `169.254.169.254` (the original repro) | 403 |
| Loopback | `127.0.0.1`, `localhost`, `[::1]` | 403 |
| RFC 1918 private | `10.0.0.1`, `192.168.1.1` | 403 |
| Typosquats | `sum.golang.org.evil.com` (suffix), `evil.com.sum.golang.org` (prefix) | 403 |
| External | `example.com` | 403 |
| **Positive control** | `sum.golang.org` | NOT 403 |
| **Case-insensitive** | `SUM.GOLANG.ORG` | NOT 403 |

Each blocked probe MUST 403 specifically (not 404, not 401). The fix's contract is explicit: 403 from the allowlist gate. Accepting 4xx-generally would let a future regression (router 404 from a typo, auth 401 middleware change) silently pass through this guard.

The positive controls assert "not 403" rather than "200" because the runner may have no internet access to sum.golang.org. What we pin is that the allowlist check passed and the request was attempted upstream — the allowlist-deny semantics, not the upstream-fetch semantics.

## Lessons applied from prior judge passes

- Server-side cleanup via `add_exit_handler` immediately after `create_local_repo` succeeds (from #183 judge pass)
- Pure URL probes, no upload — isolates the regression class (from #180 calibration learning)
- 403-specifically pin, not "any 4xx" (from #180 reality-checker note about loose pins)

## Verification

- [x] `bash -n` clean
- [ ] Local docker-compose exec deferred (backend in docker-desktop-extension network with internal-only port mapping; release-gate `workflow_dispatch` is the authoritative execution path, tracked via [#179](https://github.com/artifact-keeper/artifact-keeper-test/issues/179))

## Refs

- artifact-keeper#879 `3e529938` — the fix being guarded
- [#180](https://github.com/artifact-keeper/artifact-keeper-test/pull/180) — calibration
- [#183](https://github.com/artifact-keeper/artifact-keeper-test/pull/183) — sibling helm-based test in this batch
- [`docs/triage-2026-05-17-epic-coverage.md`](https://github.com/artifact-keeper/artifact-keeper-test/blob/main/docs/triage-2026-05-17-epic-coverage.md) §4 (untested fixes)

### DIFF
--- a/tests/security/test-goproxy-sumdb-ssrf-allowlist.sh
+++ b/tests/security/test-goproxy-sumdb-ssrf-allowlist.sh
@@ -1,0 +1,195 @@
+#!/usr/bin/env bash
+# test-goproxy-sumdb-ssrf-allowlist.sh
+#
+# Closes a sub-task of the artifact-keeper-test#181-cohort of untested
+# security fixes (triage doc §4, "Critical" tier).
+#
+# Bug guarded against
+# -------------------
+# artifact-keeper#879 (commit 3e529938, merged 2026-04-25). The Go
+# proxy handler's sumdb route forwards requests to
+# `https://{host}/{path}` where `{host}` is a URL path component
+# controlled entirely by the caller. Pre-fix, ANY host string was
+# accepted, including
+#
+#   GET /go/{repo}/sumdb/169.254.169.254/latest/meta-data/iam/...
+#
+# which made the server fetch AWS / GCP cloud-metadata on the
+# attacker's behalf and return the response body. Textbook SSRF.
+#
+# Fix: a static allowlist of permitted upstream sumdb hosts
+# (sum.golang.org, sum.golang.google.cn). Any other host returns
+# HTTP 403, comparison case-insensitive per RFC 1035.
+#
+# What this test pins
+# -------------------
+# 1. Cloud-metadata addresses (169.254.169.254) MUST be rejected.
+# 2. Loopback addresses (127.0.0.1, localhost, [::1]) MUST be rejected.
+# 3. RFC 1918 private addresses MUST be rejected.
+# 4. Typosquatted hosts (sum.golang.org.evil.com) MUST be rejected.
+# 5. Random external hosts MUST be rejected.
+# 6. Case variations of allowed hosts (SUM.GOLANG.ORG) MUST NOT
+#    return 403 from the allowlist check (RFC 1035 case-insensitivity).
+#    They may still 4xx/5xx for other reasons (no real upstream, no
+#    Go module being looked up) but NOT with the allowlist 403.
+#
+# Requires: curl
+
+source "$(dirname "$0")/../lib/common.sh"
+
+begin_suite "goproxy-sumdb-ssrf-allowlist"
+auth_admin
+
+REPO_KEY="sec-sumdb-ssrf-${RUN_ID}"
+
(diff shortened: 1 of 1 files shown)


brandonrc:
Two-judge pass complete.

**Reality Checker — APPROVE**

Traced the actual handler control flow on origin/main: `handle_get` → `resolve_go_repo` → `parse_path` → `proxy_sumdb` → `is_sumdb_host_allowed` → `StatusCode::FORBIDDEN` (goproxy.rs:260). No auth gate fires on GET (`handle_get` doesn't call `require_auth_with_bearer_fallback`; only `handle_put` does). The test creates a real local Go repo via `create_local_repo`, so `resolve_go_repo` passes; the `sumdb/{host}/lookup/...` shape parses cleanly, so `parse_path` passes; control reaches the allowlist gate. A pre-fix backend would 200 with cloud-metadata body; post-fix returns 403. Real regression check.

403-specific pin matches the fix exactly (line 260 emits `FORBIDDEN`). Positive-control "not 403" cleanly isolates allowlist-deny semantics — upstream-fetch failures use `BAD_GATEWAY` (502, lines 277/293), never 403.

IPv6 `%5B::1%5D` reaches the allowlist correctly: axum's `*path` wildcard is opaque (not segment-parsed), so brackets aren't separators. Tower percent-decodes before the handler runs, so `is_sumdb_host_allowed` actually receives `[::1]` and rejects it via string-match.

**Code Reviewer — LGTM-WITH-NITS**

Three documentation clarifications applied in `13e3374`:
1. `$CURL_TIMEOUT` unquoting: added a comment explaining the multi-arg word-split requirement (common.sh:491)
2. IPv6 comment rewritten to explain *both* layers: curl's transport-level requirement for encoded brackets, AND axum's percent-decoding before the handler sees the value
3. IMDS probe path noted as historical repro string (not load-bearing for the assertion, but grep-able)

**Non-blocking judge suggestions deferred** (not applied):
- Optional `metadata.google.internal` / `metadata.azure.com` probes — mirrors the fix's own unit tests but adds little marginal coverage over the 169.254.169.254 case
- `probe_allowed` helper to mirror `probe_blocked` symmetry — cosmetic; the 2 positive-control sites that would use it have different assertions
- Trigger release-gate `workflow_dispatch` before merge to confirm runtime behavior matches the static-trace analysis — depends on #179

PR #186, opened 2026-05-19
test(security): Grype gated out of OCI scans (#1164)

brandonrc:
## Summary

Closes a sub-task of the [#181](https://github.com/artifact-keeper/artifact-keeper-test/issues/181)-cohort (untested security fixes, [triage doc](https://github.com/artifact-keeper/artifact-keeper-test/blob/main/docs/triage-2026-05-17-epic-coverage.md) §4 Critical tier).

Guards artifact-keeper#1164 (`14f59c68`, merged 2026-05-11). Pre-fix, Grype was invoked on OCI image manifests via `grype dir:` mode, walking the manifest JSON instead of layers and returning 0 findings. The original repro: `nginx:latest` reporting "0 / 201" findings instead of "200 / 201". The fix gates Grype out of OCI artifacts via `Scanner::is_applicable`.

**Pre-flight verified**: `gh pr view 1164` confirms `state=MERGED, mergedAt=2026-05-11`. This is the discipline introduced after PR #185 was closed for guarding unmerged code (#1236).

## What the test pins

`scan_results` rows carry a `scan_type` field (`ScanType` enum: `dependency | image | license | malware`). Grype emits `dependency`; ImageScanner emits `image`. Post-fix an OCI manifest artifact MUST NOT have any completed scan with `scan_type=dependency`.

Pre-fix signature: a `scan_results` row with `scan_type=dependency`, `total_findings=0`, attributable to Grype silently producing "no findings" against manifest JSON. Its presence on this artifact fails the gate.

## Test flow

1. Create local docker (OCI) repo
2. Mint a `/v2` registry token
3. Chunked-upload config blob + layer blob (manifest PUT validates digest+size against stored blobs)
4. PUT manifest at `v2/{image}/manifests/{tag}`
5. Resolve `artifact_id` via `/api/v1/repositories/{key}/artifacts` (handles both `items[]` and `artifacts[]` shapes defensively)
6. Trigger scan via `POST /api/v1/security/scan` with `artifact_id`
7. Poll `/api/v1/security/artifacts/{id}/scans` for completed status (timeout `SCAN_TIMEOUT`, default 60s)
8. **Assert**: among completed scans for this artifact, ZERO have `scan_type=dependency`

## Skip semantics

If no scan completes within timeout, the suite SKIPS rather than passes. A backend with no scanners enabled cannot distinguish "Grype correctly gated out" from "no scanners at all"; a vacuous pass would mask a regression. Skip is explicit so the release-gate dashboard surfaces "scanner not enabled" separately from "gate verified".

## Hygiene (lessons from prior PRs)

- **Pre-flight**: `gh pr view` confirmed merged before writing (#185 mistake not repeated)
- Repo cleanup via `add_exit_handler` immediately after `create_local_repo` (#183 judge pass)
- `require_cmd curl + jq` up front
- Defensive `items[] || artifacts[]` shape handling for artifact_id resolution

## Verification

- [x] `bash -n` clean
- [ ] Local docker-compose exec deferred (backend in docker-desktop-extension network with internal-only port mapping; release-gate `workflow_dispatch` is the authoritative execution path, [#179](https://github.com/artifact-keeper/artifact-keeper-test/issues/179))

## Refs

- artifact-keeper#1164 `14f59c68` — fix being guarded
- [#180](https://github.com/artifact-keeper/artifact-keeper-test/pull/180) / [#183](https://github.com/artifact-keeper/artifact-keeper-test/pull/183) / [#184](https://github.com/artifact-keeper/artifact-keeper-test/pull/184) — prior worker+judge PRs
- Closed [#185](https://github.com/artifact-keeper/artifact-keeper-test/pull/185) — lesson learned re pre-flight
- [`docs/triage-2026-05-17-epic-coverage.md`](https://github.com/artifact-keeper/artifact-keeper-test/blob/main/docs/triage-2026-05-17-epic-coverage.md) §4


brandonrc:
Two-judge pass complete. **Reality Checker caught a critical fantasy-pass; fixed in `a395e54`.**

## Reality Checker — NEEDS WORK (correctly!)

The original assertion (`scan_type == "dependency"`) did **NOT** detect a Grype regression on OCI manifests:

- `GrypeScanner::scan_type()` returns the literal string **`"grype"`** (backend/src/services/grype_scanner.rs:143-145, verified on origin/main).
- `DependencyScanner` is a **separate** scanner (parses package.json / Cargo.toml / etc.) and emits `scan_type="dependency"` (scanner_service.rs:1070-1072).
- Pre-fix, Grype wrote `scan_type="grype"` rows on OCI artifacts. The `"dependency"` needle would never have fired — the test would have passed green even against a fully-reverted #1164.

This is the **third significant judge save** in Phase 2:
- #180 calibration: 2 of 3 tests were fantasy-passes
- #185: closed (target fix not yet on main)
- #186 (this): wrong identifier string

Fixed by changing the needle from `dependency` → `grype` (the actual string `GrypeScanner` emits). The fix's contract: post-#1164, `is_applicable` returns false for OCI so the scanner pipeline never invokes Grype, so **no `grype`-typed row** can exist for an OCI artifact. That's now what the test pins.

Other Reality Checker findings:
- `TriggerScanRequest` payload shape `{artifact_id}` confirmed correct
- `ScanListResponse.items[]` is the canonical shape; `.artifacts[]` fallback was dead code (left in defensively, harmless)
- `handle_put_manifest` does **not** validate config/layer digest+size against stored blobs — my "validates digest+size" comment was wrong. Corrected the docstring.

## Code Reviewer — LGTM-WITH-NITS (all applied)

1. `upload_blob` rewrite: write file first, derive `digest` + `size` from disk via `shasum -a 256` + `wc -c`. Removes the `printf %s + ${#content}` footgun (literal `%` interpreted as format specifier; multi-byte UTF-8 desyncs character-count from byte-count).
2. `artifact_id` resolution now polls up to 15s (race between manifest PUT return and artifacts-table row commit — same shape as `test-scan-completes.sh:295-329`).
3. Shellcheck `disable=SC2086` comments on three `$CURL_TIMEOUT`-unquoted curl invocations, matching workspace house style.
4. Malformed-JSON diagnostic on the poll loop: logs first 200 bytes of body so a backend regression emitting HTML doesn't silently look like "completed_count=0".

## Suggestions deferred (not applied)

- Positive control (assert at least one `scan_type="image"` from ImageScanner) — would tighten the test but introduces dependency on Trivy being enabled; the current SKIP path already handles "no scanners wired up". Leaving as a follow-up.

brandonrc:
Rebased onto main (was 40 commits behind) and merging — the rebase was trivial, as the
PR adds exactly one new file and conflicts with nothing.

Keeping this rather than folding it into the supply-chain tier: it is a valid
regression guard for a merged backend fix (#1164), the corpus security suite is its
natural home, and it is small. `scripts/run-suite.sh:60` globs
`tests/<suite>/**/test-*.sh`, so it is picked up automatically with no wiring — which
is the opposite of the 57 unwired format scripts problem tracked on #70/#72/#257.

Two honest caveats for the record:

1. **It first executes at the next release-gate run.** The gate does not run on PRs,
   so a green PR here is not evidence the script passes. If it needs calibration,
   that will surface then.
2. **It lands non-gating.** `security-tests` is `continue-on-error: true` and
   excluded from the gate rollup (release-gate.yml:2027-2028), so this guards by
   producing signal, not by failing the build. Scanner-routing would gate properly as
   an assertion in the supply-chain tier (next to scanner-efficacy) — worth doing when
   that tier absorbs the security suite, per ARCHITECTURE §6.

Source: backlog triage vs the Deployment Test Framework, 2026-07-16.

PR #202, opened 2026-05-27
fix(release-gate): enable DTrack in test-full and wire API key into security-tests (closes #200)

### LINKED ISSUE #200: Provision DependencyTrack for gate coverage (~3.6k LoC integration, currently unexercised)
## Summary

`helm/values-test-full.yaml` deliberately omits DependencyTrack with the comment "requires complex bootstrapping, handled separately". The `security-tests` suite includes a DTrack integration subsuite that correctly hard-fails under `RELEASE_GATE=1`:

```
FAIL: skip_suite called with RELEASE_GATE=1
      (reason: DEPENDENCY_TRACK_API_KEY and/or DEPENDENCY_TRACK_URL not set;
       DTrack integration not exercised)
a skipped suite in release-gate is silent-success; failing the gate
```

The fail-on-skip behavior is correct -- it catches the "silent-success class" that release-gate exists to prevent.

## Why we shouldn't just remove the subsuite

DTrack is not optional surface in the backend. It's a 3,190 LoC integration:

- `backend/src/services/dependency_track_service.rs` -- 2,489 lines
- `backend/src/api/handlers/dependency_track.rs` -- 701 lines
- `artifact-keeper-iac/charts/artifact-keeper/templates/dtrack-deployment.yaml` -- 323-line subchart
- Default `dependencyTrack.enabled: true` in `charts/artifact-keeper/values.yaml:494`

Removing the subsuite would leave the entire DTrack integration with zero release-gate coverage.

## Fix: stand up DTrack in the security-tests deploy

Infrastructure already exists in this repo for the `clean-install-smoke-with-deps` job:

- `helm/values-smoke-with-deps.yaml:159-167` enables DTrack with right-sized resources (2 CPU / 4 Gi).
- The chart's DTrack subchart includes a bootstrap init container that creates the admin API key.

Work needed:

1. **values-test-full.yaml** -- enable `dependencyTrack.enabled: true` with the same resource limits as `values-smoke-with-deps.yaml`.
2. **scripts/create-test-namespace.sh** -- after install completes, surface the bootstrap-generated API key from the K8s Secret (whatever name the subchart uses) and the DTrack service URL.
3. **.github/workflows/release-gate.yml** (security-tests job, around lines 949-994) -- export `DEPENDENCY_TRACK_URL=http://artifact-keeper-dtrack.${NAMESPACE}.svc.cluster.local:8080` and `DEPENDENCY_TRACK_API_KEY=<from secret>` into the job env.
4. **Namespace ResourceQuota** -- current ARC quota is 4 CPU / 8 Gi; security stack with DTrack pushes to ~7.5 CPU / 12.4 Gi. Either raise the namespace quota or route this job to the `ak-beefy-runners` pool referenced in `values-smoke-with-deps.yaml:48`.

## Effort and risk

- ~half day to wire and test
- Risk: capacity headroom on ARC runners. Mitigation listed under (4).

## Acceptance

`security-tests` DTrack subsuite passes under RELEASE_GATE=1, and the rest of `security-tests` remains green (no regression from added DTrack startup time).

brandonrc:
## Summary

The security-tests DTrack subsuite was hard-failing under `RELEASE_GATE=1` because `DEPENDENCY_TRACK_API_KEY` and `DEPENDENCY_TRACK_URL` were never set in the test cluster. The fail-on-skip behavior is correct (it catches silent-success regressions); the actual gap was that DTrack was deliberately omitted from `values-test-full.yaml` with the comment "requires complex bootstrapping, handled separately".

This PR closes that gap. DTrack is a 3,190 LoC backend integration (`services/dependency_track_service.rs` + handler + chart subchart); leaving it without release-gate coverage means SBOM upload + findings sync regressions ship undetected.

### `helm/values-test-full.yaml`

Enable DependencyTrack with the same sizing as `values-smoke-with-deps.yaml` (250m/1Gi requests, 2 CPU/4Gi limits). The chart's bootstrap init container in `artifact-keeper-iac` already handles the admin force-password-change + team API key provisioning and writes the key to `/shared/dtrack-api-key` on the backend pod. Just need a non-empty `adminPassword` to unblock the flow.

### `.github/workflows/release-gate.yml` security-tests job

New step `Wait for Dependency-Track bootstrap and extract API key` runs between the existing runner-pod-IP resolution and `Run security tests`:

1. Wait for the DTrack deployment to become Available (5 min budget; DTrack JVM startup is slow).
2. Poll the backend pod for `/shared/dtrack-api-key` via `kubectl exec` (up to 3 min).
3. Export `DEPENDENCY_TRACK_URL` and `DEPENDENCY_TRACK_API_KEY` into the job env. The key is `::add-mask::`'d so it stays out of logs.

On bootstrap failure the step warns and exits 0 (the rest of the security suite still runs). The DTrack subsuite then surfaces the bootstrap issue as a clear skip-fail rather than masking it.

## Resource impact

Adds ~2 CPU / 4 Gi memory to the test-full deploy (limits). Combined stack ~7 CPU / 10 Gi. The ARC namespace has no enforced ResourceQuota (verified: `create-test-namespace.sh` does not create one, and no chart template defines one). Cluster headroom on Rocky is adequate; if this ever changes, the right answer is to route security-tests to the `ak-beefy-runners` pool already configured in `artifact-keeper-iac/argocd/arc-beefy-runners-values.yaml`.

## Why not just remove the DTrack subsuite

It would leave a 3,190 LoC backend integration with zero gate coverage. `clean-install-smoke-with-deps` covers chart-wiring (does DTrack reach Ready), not actual SBOM upload + findings sync. The latter is what regresses silently when DTrack integration code drifts.

## Test plan

- [x] YAML validates for both files
- [ ] Manual workflow_dispatch of Release Gate after merge; verify the new step extracts the API key and the DTrack subsuite either passes or fails on real DTrack integration findings (not on missing env vars)
- [ ] Cluster headroom holds (no DTrack pod stuck in `Pending` due to insufficient resources)

Closes #200


brandonrc:
Investigation update from a focused triage on the self-test failures:

## The DTrack standup is NOT the cause of the self-test failures

The agent that dug into this confirmed DTrack does not write `scan_results` rows. It only consumes SBOMs as a one-way push (`backend/src/services/scanner_service.rs:2965`, `submit_sbom_to_dependency_track`). Enabling DTrack adds zero findings to the backend's scan-completion query.

The actual scanners that produce findings are hardcoded at `scanner_service.rs:2312-2389`:
- `DependencyScanner` (always on, queries OSV + GitHub Advisory independently)
- `ImageScanner` + `TrivyFsScanner` + `IncusScanner` (when `TRIVY_URL` is set)
- `GrypeScanner` (always on, CLI-based, parses `package-lock.json` locally)
- `OpenScapScanner` (when `OPENSCAP_URL` is set)

For **fixture A (Trivy scaled to zero)**: `DependencyScanner` + `GrypeScanner` still run unconditionally and find CVE-2019-10744 on lodash 4.17.4 via OSV/GHSA without needing the Trivy pod. The gate's `findings_count >= 1` assertion is satisfied.

For **fixture B (clean lodash 4.17.21)**: same scanners fire. The gate polls `GET /security/artifacts/{id}/scans` which returns rows from all scanners ordered `created_at DESC` and takes `.items[0]`. Whichever scanner finished most recently wins, and at least one of them is reporting findings the test doesn't expect.

So the self-test contract was broken BEFORE my DTrack changes. The failures here surface because of background scanner behavior, not DTrack.

## The right fix (separate from this PR)

The self-tests need to either:
1. Pin the gate assertion to a specific `scan_type=trivy` so non-Trivy findings don't satisfy it
2. Disable `DependencyScanner` + `GrypeScanner` in the fixture overlays (env vars or chart toggles)
3. Use a fresh in-house fixture that has no known CVEs in any of OSV/GHSA/Trivy/Grype

This is independent from this PR's scope.

## Proposed path for THIS PR

Two options:

**Option A (recommended): split DTrack out of `values-test-full.yaml`.** Move the DTrack enablement to a new `helm/values-test-dtrack.yaml` and have only the security-tests job apply it via `--values` (or `helm upgrade --reuse-values --values`). The self-tests stay on the lean `values-test-full.yaml` and remain green.

**Option B: keep this PR as-is, file a separate issue for the self-test contract.** Self-tests stay red until that lands, but the gate-side benefit of this PR (real DTrack release-gate coverage) materializes immediately.

I'd vote A for cleaner separation. Will rework if you concur.

brandonrc:
Triage note — **leaving this open as an owner decision. Not deciding it here, and not
closing it.**

This PR works. The question is not whether it is correct but whether the k8s gate is
where DTrack coverage should live, given that behavioral suites are migrating to the
DTF (ARCHITECTURE §6).

**The trade:**

| | PR #202 (k8s gate) | `integrations.dtrack` overlay (DTF) |
|---|---|---|
| Available | now, works today | needs building |
| Cost | permanent ~7.5 CPU / 12.4 Gi ARC quota bump | ~2 CPU / 4 Gi absorbed by a compose slot |
| Direction | deepens the gate suites are moving off | lands where supply-chain coverage is heading |

**The triage recommends the overlay**, on the reasoning that the quota bump is a
permanent cost paid on infrastructure with a declining lifespan, and the supply-chain
tier is the natural neighbor (scanner-routing and scanner-efficacy already live
there).

**But the counter-argument is real:** ~3.6k LoC of DependencyTrack integration
(`dependency_track_service.rs` 2,912 + `dependency_track.rs` 701) has zero exercised
coverage right now, and this PR fixes that today. If k8s-gate DTrack coverage is
wanted *this quarter*, merging this is defensible — "unexercised now" beats
"well-placed later" if the overlay is not imminent.

Either way the gap tracked in #200 stays open until one of the two lands.

Owner's call. Source: backlog triage vs the Deployment Test Framework, 2026-07-16.

PR #246, opened 2026-07-12
test(security): assert write-scope 403 on the eight GHSA-vvc3 deferred format handlers

brandonrc:
## Summary

Paired regression coverage for artifact-keeper #2417, which completes the
GHSA-vvc3-h39c-mrq5 scope-enforcement rollout on the eight previously-deferred
format publish handlers (ansible, chef, cocoapods, jetbrains, pub, puppet, sbt,
vscode).

- Removes `tests/security/test-ghsa-vvc3-deferred-handlers.sh`, which pinned the
  *still-vulnerable* state ("read-scope SA token is NOT blocked by scope on these
  eight publish paths"). That contract is now obsolete.
- Folds the eight formats into the protected assertions of
  `tests/security/test-ghsa-vvc3-scope-checks.sh`: each now asserts the publish
  path returns `403` with body exactly `Token does not have required scope: write`
  for a read-scope service-account token, identical to the existing npm / pypi /
  oci assertions. The pub upload-URL preflight (`GET .../versions/new`) is asserted
  too.

Implementation notes baked into the test:

- **sbt** is routed under `/ivy/{repo}/...` (not `/sbt`) with `PUT`; hitting `/sbt`
  would `404` and mask the scope check.
- Per-format content types are preserved. The multipart-extractor handlers
  (ansible, chef, pub, puppet) are sent with a valid multipart boundary (`curl -F`)
  so the extractor does not reject the request before the scope gate runs; the
  scope check fires ahead of any body parsing, so a minimal body suffices.
- Repos are created `is_public:true` (via `create_local_repo`) so the read-scope SA
  can resolve the repo and reach the handler's scope gate — a private repo would
  `404` first and vacuously "pass".

Verified locally against a backend built from the #2417 branch: the suite reports
33 passed / 0 failed, including all eight new format assertions and the pub
preflight.

Pairs with: artifact-keeper #2417, GHSA-vvc3-h39c-mrq5.


brandonrc:
Extended this branch with token-exchange scope-laundering coverage for artifact-keeper #2430 (v1.5.2). Four new cases assert that a read-only API token exchanged into a JWT (Conan `users/authenticate`) or an OCI bearer (docker-login `GET /v2/token`) inherits the token's action-scope ceiling and is rejected (403 `Token does not have required scope: write`) on a subsequent OCI write, with read+write positive controls (202). Verified against a v1.5.2 candidate backend carrying both #2417 and #2430: 38 passed / 0 failed.

